PR #2, opened 2023-01-30
Use default github token in actions

stellarhoof:
It's setup to use my own personal access token right now

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -2,6 +2,11 @@ name: Lint, Test, and Danger
 
 on: pull_request
 
+# https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
+permissions:
+  checks: write
+  pull-requests: write
+
 jobs:
   lint-test-danger:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -2,14 +2,19 @@ name: Lint, Test, and Danger
...
+
 jobs:
   lint-test-danger:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
         with:
-          token: ${{ secrets.PAT_DECRAPIFIER }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
rudolph9, line 17: This is going to prevent duti from pushing up commits when finds automatic formatting or lint fixes but doesn't look like we're running those in CI right now anyway.

rudolph9 (approved):

.github/workflows/pull-request.yaml
stellarhoof: Yeah we're using other actions for those. Also, we can still use this token if we give it enough permissions I think. There should be nothing special about `PAT_DECRAPIFIER` that this one can't do.

PR #3, opened 2023-01-31
Fiscal Year Pivot Click Through Fix: GS-3431

positiveVibes4all:
## Pivot Chart Intersection Selection Filtering Fix
---
- Fixed fiscal year click through and added hoisting the the filters.
- Refactored code that was impacted.


daedalus28 (changes requested):
Based on the feedback in the original PR, the refactoring here should really move hoisting to a new pivot type function for hoisting so we can get rid of the omit juggling.

The diff here also seems very large - we should probably split the bug fix and refactoring into two PRs (and judging from this diff, I think perhaps we should start with the refactor)

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -137,10 +146,10 @@ let createPivotScope = (node, schema, getStats) => {
       _.flatMap('loaded')
     )(expansions)
 
-  let isFirstOfType = (expansion) =>
+  let isFirstOfType = expansion =>
```
daedalus28: @stellarhoof these seems like prettier diffs - we should have one config for the monorepo and run it for everything in its own PR

stellarhoof: We do, but nothing's running in CI because we need to merge https://github.com/smartprocure/contexture/pull/2 since I deleted my personal token from this repo.

rudolph9 (approved):

.gitignore
```diff
@@ -18,3 +18,5 @@ coverage
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.vscode/
```
rudolph9: Can you put this in your a local only file to ignore like `.git/info/exclude` or a global gitignore https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

rudolph9 (changes requested):

.gitignore
positiveVibes4all: Sure thing

positiveVibes4all:
Another opened for this in favor of reducing commits and removing refactor work not related to this commit.

PR #6, opened 2023-01-31
Fix contexture repo links

stellarhoof:
Point to new packages and while we're at it also point to `main` (all hail the new master!)


packages/react/README.md
```diff
@@ -50,7 +50,7 @@ Documentation lives in our [Storybook](https://smartprocure.github.io/contexture
 
 - [Theming documentation](https://smartprocure.github.io/contexture-react/?path=/story/developer-notes-docs--theme-api)
 - A [Live Demo](https://smartprocure.github.io/contexture-react/?path=/story/live-demos-imdb-search--dynamic-search-layout) using IMDB data
```
daedalus28, line 52: These are interesting - how are we going to handle per package sites? I guess we can try to put it all in one place in one large site, but storybook is really kind of its own thing 🤔

daedalus28 (changes requested):
I think we also need to bump all the packages so npm updates

packages/react/README.md
stellarhoof: We can deploy under a path in the site for this repo. Also related: https://github.com/github/pages-gem/issues/724

stellarhoof: https://github.com/github/pages-gem/issues/724

stellarhoof:
I'll add changesets first and test-drive it by bumping patch versions here

daedalus28 (approved):

PR #8, opened 2023-01-31
Fix github token permissions

stellarhoof:
Give the token permissions to push and purposely add a blank line to trigger prettier committing to the repo.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,6 +5,7 @@ on: pull_request
 # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
 permissions:
   checks: write
+  contents: write
   pull-requests: write
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -5,6 +5,7 @@ on: pull_request
 # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
 permissions:
   checks: write
+  contents: write
```
rudolph9, line 8: I didn't realize token permissions are finally a thing, sick!

rudolph9 (approved):

.github/workflows/pull-request.yaml
stellarhoof: Yup. `GITHUB_TOKEN` has everything by default, and `permissions` gives it only the ones we define. We can now get rid of `PAT_DECRAPIFIER` in other places if we want to. Another cool thing is that we can give permissions to the token at the job level instead of the workflow level: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

PR #10, opened 2023-02-02
Use yarn workspaces for intra-package dependencies

stellarhoof:
This will ensure we're on the same version of packages for development


stellarhoof:
Closing. This makes bumping versions confusing.

PR #12, opened 2023-02-08
Hoist Properties Refactor and Fix Clickthrough

positiveVibes4all:
- Refactor old code to use new pivot type function instead
- Added support for meta filters to have associated hoist data for click through fix


rudolph9 (approved):

re1ro (approved):
Does it make sense to add some tests to cover new functions?

positiveVibes4all:
> 

Yes, I have the core tests still in the last commit for this still as well. I will also add some tests like drilldown with various combinations at different levels in a subsequent PR.

packages/provider-elasticsearch/src/index.js
```diff
@@ -37,7 +37,21 @@ let ElasticsearchProvider = (config = { request: {} }) => ({
...
+  async runSearch(
+    { requestOptions = {} } = {},
+    node,
+    schema,
+    filterData,
+    aggData
+  ) {
+    //Unpack filters and query data if hoisted data present, or use regular values
+    let aggs = aggData && aggData.hoistProps ? aggData.aggs : aggData
+    let filters =
```
daedalus28: This is still the same kind of stuff we were doing before with omit. Lets not complicate things like this if we can avoid it.
In fact, do we even need to hoist to the root? The prior implementation hoisted to a parent agg level which @rudolph9 thought might be more efficient

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -210,6 +210,55 @@ let createPivotScope = (node, schema, getStats) => {
...
+   *   This is to avoid having issues in which props are not allowed at the same level.
+   *   hoistProps allows groups to hoist items to top of mapping structure and can be
+   *   used for other needs in which hoisting is required.
+   */
+  let getHoistProps = async ({
+    drilldown = { rowDrills: [], columnDrills: [] },
+    groups = { rows: [], columns: [] },
+  }) =>
+    _.reduce(
+      _.merge,
```
daedalus28: This is probably simpler with mergeAll - there’s a lot of nesting here that we can reduce by expressing this as a flow

daedalus28 (changes requested):

packages/provider-elasticsearch/src/index.js
positiveVibes4all: This came about due to when you select intersections and click show results, the meta filters encapsulate the filters from the report node on its run. In this case the filters are brought in without the aggs. So in this way if a hoisted property is needed it has to be above all of the filters for it to work properly.

I will look into just adding the hoisting without the unpacking in results function, but I think the filters were the problematic part here when running on a result node from a report node selection.

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
positiveVibes4all: I will rewrite this, but the new prettier configs are exploding the code more than it used to also.  

Edit: getDrilldown is really the only thing marrying values to groups, but since it is duplicated code it might help them both to make a function that pairs filter values with the groups.

packages/provider-elasticsearch/src/index.js
positiveVibes4all: For example when you click through to results tab and filters are generated from the report groups this is what is generated currently, you can use this to see where it is legal to put runtime mappings in this given case to help guide our discussion around this:

Using `govspend_beta.federalombittowers/_search`:

https://gist.github.com/positiveVibes4all/104da9fd38fb2a8b213e894aa62ce0f6

daedalus28: If we need to hoist to the root, that's fine - but juggling like this seems like incidental complexity. Ideally we can just call something to get hoisted values directly, and not sometimes pack them up in filters/aggs and conditionally unpack.

positiveVibes4all: I should be able to add a step in the server project, but that would add to the documented flow so it seemed like a bigger impact than just getting it working in the pivot case first as it only impacts the elastic provider. 

I can rewrite the server to attachHoisted props and pass to the curriedSearch like relevant Filters I just need to work out if I could do the same with the aggs as well which I think I could with minimal impact, but this will also have a bigger impact over the whole code so I was trying to avoid that. Maybe we can meet up for a little to talk about it?

positiveVibes4all: Also the getHoistedProps does get them directly now from the groups with no need to pass anything around, it really just comes down to needing to persist them when relevant Filters include the aggregation nodes as the source of the filters in the server flow.

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
positiveVibes4all: I implemented the feedback here in the latest PR

packages/provider-elasticsearch/src/index.js
positiveVibes4all: I addressed this issue in the latest PR, please let me know your thoughts.

positiveVibes4all:
I need to update the tests still, but it would still be good to get the feedback on the PR as soon as possible while I work on that. The test impact is pretty significant, because they all have an added property they are not expecting, I will hold off on that commit until this one is looked over to avoid a huge delta while reviewing.

re1ro (approved):

re1ro (approved):

daedalus28:
Just an update here - we're going to go a much simpler route using some newer futil methods to just unset and collect hoisting in runSearch off filters and aggregations so we can remove all the plumbing.

daedalus28:
Just some additional notes:

At some point we'll like add a transduce method to either futil trees or futil more broadly, but for now:

```js
// For futil, but can live in this PR:
let unsetOnTree = _.curry((prop, tree) =>
  _.compact(Tree.toArrayBy(F.popProperty(prop))(tree))
)

// This is the method that will rip off and collect all hoisted props in an array 
export let hoist = unsetOnTree('__hoistProps')
```

Here's the test case I had locally to show how hoist works:

```js
describe('hosting', () => {
  it('should hoist', () => {
    let input = {
      aggs: {
        groups: {
          date_histogram: {
            field: 'PO.IssuedDate.fiscal',
            interval: 'year',
            min_doc_count: 0,
            __hoistProps: {
              runtime_mappings: {
                'PO.IssuedDate.fiscal': {
                  script: {
                    params: { monthOffset: 3 },
                    source: `if(doc['PO.IssuedDate'].size()!=0){${''}emit(doc['PO.IssuedDate']${''}.value.plusMonths(params['monthOffset']).toInstant().toEpochMilli())}${''}`,
                  },
                  type: 'date',
                },
              },
            },
          },
          aggs: {
            min: { min: { field: 'LineItem.TotalPrice' } },
            max: { max: { field: 'LineItem.TotalPrice' } },
            avg: { avg: { field: 'LineItem.TotalPrice' } },
            sum: {
              sum: {
                field: 'LineItem.TotalPrice',
                __hoistProps: {
                  runtime_mappings: {
                    'PO.OtherDate.fiscal': {
                      script: {
                        params: { monthOffset: 3 },
                        source: `if(doc['PO.OtherDate'].size()!=0){${''}emit(doc['PO.OtherDate']${''}.value.plusMonths(params['monthOffset']).toInstant().toEpochMilli())}${''}`,
                      },
                      type: 'date',
                    },
                  },
                },
              },
            },
          },
        },
      },
    }

    let output = {
      result: {
        aggs: {
          groups: {
            date_histogram: {
              field: 'PO.IssuedDate.fiscal',
              interval: 'year',
              min_doc_count: 0,
            },
            aggs: {
              min: { min: { field: 'LineItem.TotalPrice' } },
              max: { max: { field: 'LineItem.TotalPrice' } },
              avg: { avg: { field: 'LineItem.TotalPrice' } },
              sum: { sum: { field: 'LineItem.TotalPrice' } },
            },
          },
        },
      },
      removed: [
        {
          runtime_mappings: {
            'PO.IssuedDate.fiscal': {
              script: {
                params: { monthOffset: 3 },
                source: `if(doc['PO.IssuedDate'].size()!=0){${''}emit(doc['PO.IssuedDate']${''}.value.plusMonths(params['monthOffset']).toInstant().toEpochMilli())}${''}`,
              },
              type: 'date',
            },
          },
        },
        {
          runtime_mappings: {
            'PO.OtherDate.fiscal': {
              script: {
                params: { monthOffset: 3 },
                source: `if(doc['PO.OtherDate'].size()!=0){${''}emit(doc['PO.OtherDate']${''}.value.plusMonths(params['monthOffset']).toInstant().toEpochMilli())}${''}`,
              },
              type: 'date',
            },
          },
        },
      ],
    }
    let result = { result: input, removed: hoist(input) }
    expect(result).toEqual(output)
  })
})
```

packages/provider-elasticsearch/src/utils/results.js
```diff
@@ -0,0 +1,4 @@
+import { unsetOnTree } from './futil.js'
+
+// This is the method that will rip off and collect all hoisted props in an array 
+export let hoistOnTree = unsetOnTree('__hoistProps')
```
rudolph9: no new line at the end of this file.

rudolph9 (commented):
looks good other than the new line thing

packages/provider-elasticsearch/src/utils/results.js
rudolph9: Good explanation: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

positiveVibes4all: Our pipeline added the new line back to the end of the file already

PR #33, opened 2023-03-09
Add ability to hide label counts in ExpandableTagsQuery

stellarhoof:
☝️


stellarhoof:
Closing since it's not needed anymore. The counts can be hidden by using `forceFilterOnly: true` on the node

PR #35, opened 2023-03-18
Make greyVest the default theme

stellarhoof:
Remove blueberry, material, and merge base theme into grey-vest

Had to break down ./src/utils/theme.js into multiple files because of
circular imports


daedalus28:
Closing for now - we'll likely eventually drop theming, but not until contexture-react becomes sufficiently trivial (by moving logic into the client) that creating alternate versions makes sense.

PR #41, opened 2023-03-26
Bugfix

daedalus28:
Make markedForUpdateChecks more defensive


daedalus28:
Turns out this was likely not the issue, this was: https://github.com/smartprocure/contexture/pull/43

PR #45, opened 2023-03-28
Fix console errors

daedalus28:
[Omit non native props to fix ExpandableTagsQuery console errors](https://github.com/smartprocure/contexture/commit/9035b4e14e24fd67639afbfe14b9f5e422d7c410)


daedalus28:
I think this was introduced here: https://github.com/smartprocure/contexture/commit/a4e25d6d3ec9a06858b6f2b13433677e8e171183

PR #51, opened 2023-03-30
feat(client): `hasResults` prop for result and pivot nodes

re1ro:
* [x] `results` and `pivot` nodes: `hasResults` prop 
* [x] Refacor: removing `mobx` `toJS` dependency
* [x] Refacor: passing snapshot through `runTypeFunctionOrDefault` to `init` and other node methods

### DIFF
--- a/packages/client/src/exampleTypes.js
+++ b/packages/client/src/exampleTypes.js
@@ -126,6 +126,14 @@ export default F.stampKey('type', {
     },
   },
   results: {
+    init(node, extend, snapshot) {
+      if (_.isNil(node.hasResults))
+        Object.defineProperty(node, 'hasResults', {
+          get: () => {
+            return !_.isEmpty(snapshot(node.context.results))
+          },
+        })
+    },
     validate: () => false,
     reactors: {
       page: 'self',
--- a/packages/client/src/exampleTypes/pivot.js
+++ b/packages/client/src/exampleTypes/pivot.js
@@ -1,6 +1,5 @@
 import _ from 'lodash/fp.js'
 import F from 'futil'
-import { toJS } from 'mobx'
 
 let getKey = (x) => x.keyAsString || x.key
 
@@ -16,9 +15,7 @@ let resultsForDrilldown = (type, drilldown, results) => {
 }
 
 let previouslyLoadedKeys = (expansion, expansions) => {
-  expansion = toJS(expansion)
   return _.flow(
-    toJS,
     _.filter(
       ({ type, drilldown, loaded }) =>
         type === expansion.type &&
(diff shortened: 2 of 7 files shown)


packages/client/src/serialize.js
```diff
@@ -23,7 +23,14 @@ let mapTree = (fn, tree) =>
...
 export default (tree, types, { search } = {}) => {
   let onSerialize = (node) =>
-    runTypeFunctionOrDefault(_.identity, types, 'onSerialize', node, {})
+    runTypeFunctionOrDefault(
+      _.identity,
+      types,
+      'onSerialize',
+      node,
+      {},
+      _.identity
```
daedalus28, line 32: shouldn't this be `snapshot`?

packages/client/src/index.js
```diff
@@ -115,7 +115,12 @@ export let ContextTree = _.curry(
       if (event.node)
         // not all dispatches have event.node, e.g. `refresh` with no path
         F.maybeCall(typeProp('onDispatch', event.node), event, extend)
-      await validate(runTypeFunction(types, 'validate'), extend, tree)
+      await validate(
+        runTypeFunction(types, 'validate'),
+        extend,
+        tree,
+        _.identity
```
daedalus28, line 122: shouldn't this be `snapshot`?

packages/client/src/exampleTypes/pivot.js
```diff
@@ -161,6 +109,72 @@ export let skipResetExpansionsFields = [
 ]
 
 export default {
+  init: (node, extend, snapshot) => {
+    if (_.isNil(node.hasResults))
+      Object.defineProperty(node, 'hasResults', {
```
daedalus28, line 114: Instead of adding this during init, it probably makes more sense to define it as a first class thing types can specify and have the core call `hasResults` on init and when processing responses for nodes. Using a getter is interesting since we can in theory defer computation until it's accessed - but I think in practice, it'll be used by anything observing it. Unless we're memoizing and caching the results, I think it actually ends up less performant than just calling whenever we process responses (since the code will only run once instead of on every access).

packages/client/src/exampleTypes.js
```diff
@@ -126,6 +126,14 @@ export default F.stampKey('type', {
     },
   },
   results: {
+    init(node, extend, snapshot) {
+      if (_.isNil(node.hasResults))
+        Object.defineProperty(node, 'hasResults', {
```
daedalus28, line 131: Same as below - I think this makes more sense as a new core feature that types can opt into

daedalus28 (changes requested):

re1ro:
Closing this one to split into 2 PRs

PR #53, opened 2023-03-31
Use `futil`, not `futil-js`, when importing F

dshishkov:
Use `futil`, not `futil-js`, when importing F

### DIFF
--- a/.changeset/funny-eyes-speak.md
+++ b/.changeset/funny-eyes-speak.md
@@ -1,0 +1,5 @@
+---
+'contexture-client': patch
+---
+
+Use `futil`, not `futil-js`, when importing F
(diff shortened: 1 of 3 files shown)


stellarhoof (approved):

rudolph9 (changes requested):
can you adjust this comment:

```
packages/client/README.md:The client maintains a flat tree in addition to the actual tree, which is an object mapped using `flattenTree` from `futil-js`.
````

`futil-js` -> `futil` which I guess does really matter since it' not directly saying the package name but idk do what you think is best because the repo is `smartprocure/futil-js` :shrug:

rudolph9 (approved):

PR #55, opened 2023-04-01
feat(client): `hasResults` prop for the node

re1ro:
* [x] Adding new `hasResults` prop to the node


packages/client/src/types.js
```diff
@@ -14,8 +14,10 @@ export let getTypePropOrError = _.curry(
 )
 
 export let runTypeFunctionOrDefault = _.curry(
-  (defaultFn, types, prop, node, extend) =>
-    (getTypeProp(types, prop, node) || defaultFn)(node, extend)
+  (defaultFn, types, prop, node, extend) => {
+    let typeFunction = getTypeProp(types, prop, node)
+    return (_.isFunction(typeFunction) ? typeFunction : defaultFn)(node, extend)
```
daedalus28: When would typeFunction be defined but not be a function? Is it because hasResult is the prop name and the function name? If so, it might make sense to change the name of one of them 🤔

packages/client/src/exampleTypes/pivot.js
```diff
@@ -212,6 +214,7 @@ export default {
       },
      */
     ],
+    hasResults: false,
```
daedalus28: You shouldn't have to set this in the type definition because it's in the node defaults

packages/client/src/exampleTypes.js
```diff
@@ -138,6 +138,7 @@ export default F.stampKey('type', {
       page: 1,
       pageSize: 10,
       view: 'table',
+      hasResults: false,
```
daedalus28: You shouldn't have to set this in the type definition because it's in the node defaults

daedalus28 (changes requested):

packages/client/src/exampleTypes/pivot.js
```diff
@@ -69,6 +69,8 @@ let maybeRemoveSelectedRows = (extend, node) => {
 let resetExpansions = (extend, node) => {
   extend(node, {
     expansions: [],
+    hasResults: false,
+    context: { results: {} },
```
daedalus28, line 73: Does reset expansions actually clear all results? 🤔

.changeset/giant-paws-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'contexture-client': minor
+---
+
+Added `hasResults` prop for `results` and `pivot` nodes
```
daedalus28: You also added it for every other node with a default function 👍

We should also add this to the documentation in the readme file

daedalus28 (changes requested):

packages/client/src/exampleTypes/pivot.js
re1ro: It actually should. Otherwise, we can see all results in the newly configured report. And those old results are showing when we load new results and can have wrong columns formatting etc because new pivot configuration applied while having results for the previous configuration

daedalus28 (approved):

PR #69, opened 2023-04-12
feat(elasticsearch): filterNested for highlight

re1ro:
* [x] Provider ES: `filterNested` flag for highlight config


daedalus28:
We should add test cases for this

daedalus28 (approved):

PR #71, opened 2023-05-04
Add Keyword Generation Hook

positiveVibes4all:
## Summary 
Added code to facilitate a keyword generation hook into all tags query fields  using a prop to conditionally enable. One change was made to contexture that allows for reactors to be functions and not need to map back to the predefined reaction functions.


packages/react/package.json
```diff
@@ -73,6 +73,8 @@
     "write-pkg": "^4.0.0"
   },
   "dependencies": {
+    "@heroicons/react": "1.0.6",
```
stellarhoof: Why this icons library too?

packages/react/package.json
```diff
@@ -73,6 +73,8 @@
...
+    "axios": "^1.3.6",
```
stellarhoof: Is this being used anywhere?

packages/react/package.json
```diff
@@ -81,6 +83,7 @@
     "react-date-picker": "^7.2.0",
     "react-dnd": "^2.5.4",
     "react-dnd-html5-backend": "^2.5.4",
+    "react-icons": "^4.8.0",
```
stellarhoof: Is this being used anywhere?

packages/react/src/greyVest/Tag.js
```diff
@@ -27,9 +30,10 @@ let Tag = ({
       wordBreak: 'break-all',
       ...F.callOrReturn(tagStyle, value),
     }}
-    onClick={onClick}
+    onClick={_.isFunction(moveTag) ? moveTag({ value, label }) : onClick}
```
stellarhoof: Instead of taking `moveTag`, just pass `value` and `label` to `onClick`

stellarhoof (changes requested):
@positiveVibes4all Can you add stories for this new feature

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
-          (tag) => filter({ ...node, tags: [tag] }),
-          node.tags
-        ),
+let buildResultQuery = (
+  node,
+  children = {},
+  groupsKey = 'tags',
+  keywordGenerations = []
+) => {
+  return {
```
daedalus28: You don't need a return here - this can use shorthand syntax. Our lint rules should generally warn against it, but by omitting the return, you'll also reduce the diff size

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
+        ...children,
       },
-      ...children,
+      ...(keywordGenerations.length > 0 && {
+        keywordGenerations: {
+          filters: {
+            filters: F.compactObject(
+              F.arrayToObject(
+                _.identity,
+                (word) => filter({ ...node, tags: [{ word: `${word}` }] }),
```
daedalus28: I don't think we need to wrap word in a template string

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
+        },
+        ...children,
       },
-      ...children,
+      ...(keywordGenerations.length > 0 && {
+        keywordGenerations: {
+          filters: {
+            filters: F.compactObject(
+              F.arrayToObject(
+                _.identity,
```
daedalus28: F.keysToObject does this under the hood

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
-    _.mapValues(_.get('doc_count')),
-    (results) => ({ results })
+    (results) => ({
+      tags: _.get('aggregations.tags.buckets', results),
+      keywordGenerations: _.flow(
+        _.get('aggregations.keywordGenerations.buckets'),
+        _.omitBy((x) => x.doc_count === 0)
+      )(results),
+    }),
+    F.map(F.map('doc_count'))
```
daedalus28: F.map is marked deprecated and probably not what you want 🤔

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
 
   return _.flow(
-    _.get('aggregations.tags.buckets'),
-    _.mapValues(_.get('doc_count')),
-    (results) => ({ results })
+    (results) => ({
+      tags: _.get('aggregations.tags.buckets', results),
+      keywordGenerations: _.flow(
+        _.get('aggregations.keywordGenerations.buckets'),
+        _.omitBy((x) => x.doc_count === 0)
```
daedalus28: You can invert this and use short hand syntax:

`_.pickBy('doc_count')`

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
-let result = async (node, search) => {
-  let aggs = buildResultQuery(node)
+let result = (generateKeywords) => async (node, search) => {
+  let keywords =
+    node.generateKeywords === true
+      ? await generateKeywords(generationTagInputs(node.tags))
+      : []
+  let aggs = node.generateKeywords
+    ? buildResultQuery(node, {}, 'tags', keywords)
+    : buildResultQuery(node)
```
daedalus28: This switch is weird because buildResultQuery already has full knowledge of `node`

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -80,28 +81,66 @@ let filter = ({ tags, join, field, exact }) => ({
...
-  },
-})
+  }
+}
 
-let result = async (node, search) => {
-  let aggs = buildResultQuery(node)
+let result = (generateKeywords) => async (node, search) => {
+  let keywords =
+    node.generateKeywords === true
```
daedalus28: This is equivalent to just checking node.generateKeywords since any truthy value is fine

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
daedalus28: Also - by hardcoding the aggregation key, that implies to me that you didn't test this in the pivot scenario. Its probably fine if it doesn't work there, but we should be explicit and at least add a comment about it

packages/provider-elasticsearch/src/utils/keywordGenerations.js
```diff
@@ -0,0 +1,8 @@
+import _ from 'lodash/fp.js'
+
+export let generationTagInputs = _.flow(
+  _.map('word'),
+  // Removing numbers from the list of words to generate keywords from as
+  // number throw off the keyword generation in most cases.
+  _.remove((word) => !isNaN(word) && !isNaN(parseFloat(word)))
```
daedalus28, line 7: I think the lodash util handles all cases we need

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -7,39 +7,137 @@ import ExpandArrow from './ExpandArrow.js'
...
+              moveTag={({ value, label }) =>
+                (e) => {
+                  tree.mutate(node.path, {
+                    ...(_.flow(
+                      _.find({ value: _.trim(value) }),
+                      _.isUndefined
+                    )(node.tags) && {
+                      tags: [...node.tags, { word: value, distance: 3, label }],
+                    }),
+                    context: {
```
daedalus28: Mutating context is a red flag and generally should never be done. What's the reasoning here?

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -7,39 +7,137 @@ import ExpandArrow from './ExpandArrow.js'
...
+          {_.map((word) => (
+            <theme.Tag
+              tree={tree}
+              node={node}
+              moveTag={({ value, label }) =>
+                (e) => {
+                  tree.mutate(node.path, {
+                    ...(_.flow(
+                      _.find({ value: _.trim(value) }),
+                      _.isUndefined
```
daedalus28: I think you can rely on falseyness and don't need an undefined check

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -79,73 +186,114 @@ let TagsWrapper = observer(
...
+                    await tree.mutate(node.path, { generateKeywords: true })
+                    // Await on the first generation to update false,because
+                    // subsequent button clicks(hide/show) can have edge cases
+                    tree.mutate(node.path, { generateKeywords: false })
+                  }
+                } else {
+                  if (!node.generateKeywords) {
+                    await tree.mutate(node.path, {
+                      generateKeywords: true,
+                      context: { keywordGenerations: {} },
```
daedalus28: Mutating context directly is a red flag and seems like it should live in the client somewhere

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
positiveVibes4all: it still works in the pivot scenario because it uses buildResultQuery to build it up from what I remember. I can take a look again but all of those tests are passing right now for pivot

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
positiveVibes4all: The idea here is to remove them from the context so the generations are no longer in the list.

packages/provider-elasticsearch/src/utils/keywordGenerations.js
positiveVibes4all: From what I looked into the loadash util does not accurately find numbers that are strings. There seemed to be an old method isNumeric that did this though when I looked into it. This function seems to be removed though.

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
positiveVibes4all: regular map should work, I will swap it out.

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
positiveVibes4all: no this is not needed, I will remove that and simplify it

packages/react/package.json
positiveVibes4all: No, I would up using the svg option for hero v11 icons instead because the library is a very old one. I will remove this.

packages/react/package.json
positiveVibes4all: Both of these should go away, I did not remove them.

packages/react/package.json
positiveVibes4all: I tried several ways to get the Hero V1 icons in as they are very old and do not work like the modern ones we are using currently.  This can go away also.

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
positiveVibes4all: Fixed, this was so that I could show loading stuff correctly. I am now using the node properties that reflect this state.

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
positiveVibes4all: Here I am checking if it is undefined, so we do not add duplicates in certain edge cases.

packages/react/src/greyVest/Tag.js
positiveVibes4all: Implemented this feedback

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
positiveVibes4all: swapped out for mapValues

.changeset/rich-otters-hunt.md
stellarhoof, line 1: This file should not have been deleted

positiveVibes4all: Added it back, it got removed on accident with the others.

packages/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-client",
-  "version": "2.50.2",
+  "version": "2.51.0",
```
stellarhoof: You don't need to change the version manually, changesets will take care of it

packages/provider-elasticsearch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-elasticsearch",
-  "version": "1.21.0",
+  "version": "1.22.0",
```
stellarhoof: Same

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.56.0",
+  "version": "2.57.0",
```
stellarhoof: Same

packages/react/package.json
```diff
@@ -54,8 +54,8 @@
     "@storybook/react-webpack5": "^7.0.0-alpha.0",
     "@storybook/storybook-deployer": "^2.8.16",
     "contexture": "^0.12.19",
-    "contexture-client": "^2.50.2",
-    "contexture-elasticsearch": "^1.19.16",
+    "contexture-client": "^2.51.0",
+    "contexture-elasticsearch": "^1.22.0",
```
stellarhoof: No need to bump these. Changesets will take care of it

packages/react/package.json
stellarhoof: Still here

packages/react/package.json
positiveVibes4all: these were left over from the previous bump I reverted to grab your new updates for the outside click push. Will clean it up.

packages/react/src/greyVest/Tag.js
```diff
@@ -27,9 +28,10 @@ let Tag = ({
       wordBreak: 'break-all',
       ...F.callOrReturn(tagStyle, value),
     }}
-    onClick={onClick}
+    onClick={onClick({ value, label })}
```
stellarhoof: This should be `onClick={() => onClick({ value, label })}`

positiveVibes4all: If it is a function call with no params in an event handler, should we not just use the named function though. That is what I learned a while back, and looking it up seems like it is still valid concern.

https://stackoverflow.com/questions/74434193/anonymous-function-vs-named-function-on-event-handlers

positiveVibes4all: I updated the code appropriately

packages/client/src/exampleTypes.js
```diff
@@ -104,16 +104,34 @@ export default F.stampKey('type', {
...
       exact: 'all',
+      //Only react when generateKeywords is true
+      generateKeywords: (parent, node, event) =>
+        event.value.generateKeywords ? [node] : [],
     },
     defaults: {
+      generateKeywords: false,
       field: null,
       tags: [],
+      keywordGenerations: [],
```
daedalus28: This should be in context

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -7,36 +7,142 @@ import ExpandArrow from './ExpandArrow.js'
...
+              <theme.Tag
+                tree={tree}
+                node={node}
+                onClick={({ value, label }) =>
+                  (e) => {
+                    tree.mutate(node.path, {
+                      ...(_.flow(
+                        _.find({ value: _.trim(value) }),
+                        _.isUndefined
+                      )(node.tags) && {
```
daedalus28: this flow should be unnecessary since you're deduping below

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -7,36 +7,142 @@ import ExpandArrow from './ExpandArrow.js'
...
+                label={`${word} (${toNumber(
+                  _.get(`context.keywordGenerations.${word}`)(node)
+                )})`}
+              />
+            ))(
+              _.flow(_.intersection, (dups) =>
+                _.pullAll(dups, _.keys(node.context.keywordGenerations))
+              )(
+                _.keys(node.context.keywordGenerations),
+                _.keys(node.context.tags)
```
daedalus28: This intersection flow seems like it can be a simple `_.filter`

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -7,36 +7,142 @@ import ExpandArrow from './ExpandArrow.js'
...
+                onClick={({ value, label }) =>
+                  (e) => {
+                    tree.mutate(node.path, {
+                      ...(_.flow(
+                        _.find({ value: _.trim(value) }),
+                        _.isUndefined
+                      )(node.tags) && {
+                        tags: [
+                          ...node.tags,
+                          { word: value, distance: 3, label },
```
daedalus28: We should factor out a method for this so that `distance: 3` isn't in multiple places

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -55,16 +161,26 @@ let TagsWrapper = observer(
...
           ...props,
-          ...(!_.isNil(result)
-            ? { label: `${props.value} (${toNumber(result)})` }
-            : {}),
+          ...(!_.isNil(result) && {
+            label: `${props.value} (${toNumber(result)})`,
+          }),
+          ...(_.isNil(result) &&
+            !_.isNil(generations) && {
+              label: `${props.value} (${toNumber(generations)})`,
```
daedalus28: You can simplify this by cascading the two paths:
```js
let count = F.cascade(
  [`context.tags.${props.value}`, `context.keywordGenerations.${props.value}`],
  node
)
```
and then in tagProps, just have one case where you're spreading on an object with label

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -79,73 +195,114 @@ let TagsWrapper = observer(
...
-              />
-            )}
-          </Popover>
-        </GridItem>
-      </Grid>
+              onClick={async () => {
+                // Ensure first time user clicks the lightbulb words generate.
+                // Regenerate words and reset existing for clean loading experience
+                // on subsequent clicks.
+                console.log('generationsCollapse', generationsCollapse, node)
```
daedalus28: errant console log

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -79,73 +195,114 @@ let TagsWrapper = observer(
...
+                if (F.view(generationsCollapse)) {
+                  F.off(generationsCollapse)()
+                  if (
+                    !node.context.keywordGenerations &&
+                    !node.generateKeywords
+                  ) {
+                    await tree.mutate(node.path, { generateKeywords: true })
+                    // Await on the first generation to update false,because
+                    // subsequent button clicks(hide/show) can have edge cases
+                    tree.mutate(node.path, { generateKeywords: false })
```
daedalus28: let's factor this out to a method since it's in here twice and the logic is not obvious

packages/client/src/exampleTypes.js
```diff
@@ -104,16 +104,34 @@ export default F.stampKey('type', {
...
       join: 'any',
       exact: false,
       context: {
         results: {},
       },
     },
+    onSerialize: (node) => ({
+      ..._.omit('generateKeywords', node),
+    }),
```
daedalus28: `onSerialize: _.omit('generateKeywords')`

packages/client/src/exampleTypes.js
```diff
@@ -342,5 +360,10 @@ export default F.stampKey('type', {
       searchId: null,
       search: null,
     },
+    onSerialize(node) {
+      return {
+        ..._.omit('generateKeywords', node),
+      }
+    },
```
daedalus28: this shouldn't be here

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -122,5 +157,5 @@ export default {
   filter,
   validContext,
   buildResultQuery,
-  result,
-}
+  result: result(options?.generateKeywords || (() => [])),
```
daedalus28: You've got local changes for this based on what we discussed, but the short version is to move the default logic to the place where its used (and actually just use maybeCall to not worry about it)

packages/provider-elasticsearch/src/utils/keywordGenerations.js
```diff
@@ -0,0 +1,8 @@
+import _ from 'lodash/fp.js'
+
+export let generationTagInputs = _.flow(
```
daedalus28: rename to sanitizeTagInputs

packages/provider-elasticsearch/src/utils/keywordGenerations.js
daedalus28, line 7: add test cases since this is not obvious

packages/provider-elasticsearch/test/example-types/filters/tagsQuery.test.js
```diff
@@ -13,7 +13,7 @@ let {
   filter,
   buildResultQuery,
   result,
-} = tagsQuery
+} = tagsQuery(() => [])
```
daedalus28: this invocation is wrong since tagsQuery takes an object - and with the maybeCall changes, you can just pass in nothing

packages/provider-elasticsearch/src/example-types/metricGroups/tagsQueryGroupStats.js
```diff
@@ -1,7 +1,7 @@
 import { groupStats } from './groupStatUtils.js'
 import tagsQuery from '../filters/tagsQuery.js'
 
-let { buildResultQuery, filter } = tagsQuery
+let { buildResultQuery, filter } = tagsQuery(() => [])
```
daedalus28: this invocation is wrong since tagsQuery takes an object - and with the maybeCall changes, you can just pass in nothing

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -92,16 +98,45 @@ let buildResultQuery = (node, children = {}, groupsKey = 'tags') => ({
...
+  let aggs = node.generateKeywords
+    ? buildResultQuery(node, {}, 'tags', keywords)
+    : buildResultQuery(node)
 
   return _.flow(
-    _.get('aggregations.tags.buckets'),
-    _.mapValues(_.get('doc_count')),
-    (results) => ({ results })
+    (results) => ({
+      tags: _.get('aggregations.tags.buckets', results),
```
daedalus28: assuming you can omit zero count tags and there's nothing else on `aggregations`, you can unify the code path here and remove any mention of `tags` and `keywordGenerations`

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
daedalus28: For example:

```js
_.mapValues(
  (x) => F.compactMapValues('doc_count', x.buckets),
  (await search(aggs)).aggregations
)
```

until we add `compactMapValues` to futil, you can inline it:

```js
let compactMapValues = _.flow(_.mapValues, F.compactObject)
```

daedalus28 (approved):

PR #72, opened 2023-05-08
Replace contexture-react outside click handler library

stellarhoof:
The current `react-outside-click-handler` library is 4 years old and it's not packaged right so it doesn't work in storybook. This PR replaces it with chakra's equivalent library, which uses a hook instead.


stellarhoof:
@positiveVibes4all With this you'll be able to write stories for https://github.com/smartprocure/contexture/pull/71

positiveVibes4all (approved):
Looks good, and thanks!

PR #74, opened 2023-05-12
feat(client): runTypeFunction actions

re1ro:
Originally we had `runTypeFunction` that would pass only `extend` method to type functions such as `init`, `onDispatch`, `onUpdateByOthers` etc. But to be able to work with the pivot node state we often need the node in plain JS and not MOBX observable. To accommodate that `toJS` was imported directly from `mobx` which added it as a bundled dependency. 

This PR refactors `runTypeFunction` to pass down all action functions instead of just `extend` and removes `mobx` import and bundled dependency.

* [x] client: Passing actions to the `runTypeFunction`.
* [x] client: Removing bundled `mobx` dependency

### DIFF
--- a/packages/client/src/node.js
+++ b/packages/client/src/node.js
@@ -30,16 +30,21 @@ export let internalStateKeys = [
   'onMarkForUpdate',
   'afterSearch',
   'forceReplaceResponse',
+  'expand',
+  'collapse',
 ]
 
 export let autoKey = (x) => F.compactJoin('-', [x.field, x.type]) || 'node'
 
 export let initNode = _.curry(
-  ({ extend, types, snapshot }, dedupe, parentPath, node) => {
-    runTypeFunction(types, 'init', node, extend)
+  (actions, dedupe, parentPath, node) => {
+    const { extend, types, snapshot } = actions
+    actions = _.omit(['types'], actions)
+
+    runTypeFunction(types, 'init', node, actions)
     let key = dedupe(
       node.key ||
-        runTypeFunctionOrDefault(autoKey, types, 'autoKey', node, extend)
+        runTypeFunctionOrDefault(autoKey, types, 'autoKey', node, actions)
     )
     extend(node, {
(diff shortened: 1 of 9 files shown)


packages/client/src/node.js
```diff
@@ -30,26 +30,29 @@ export let internalStateKeys = [
...
-      // For some reason, type defaults can end up observable in real world apps, so we `snapshot` instead of `_.deepClone`
-      ..._.omit(_.keys(node), snapshot(getTypeProp(types, 'defaults', node))),
-      key,
-      path: [...parentPath, key],
-    })
-  }
-)
+export let initNode = _.curry((props, dedupe, parentPath, node) => {
+  let { types, extend, snapshot } = props
+  let actionProps = _.omit(['types'], props)
```
daedalus28: Do we need to omit types? It's probably safe to pass it along 🤔

packages/client/src/serialize.js
```diff
@@ -23,7 +23,12 @@ let mapTree = (fn, tree) =>
 
 export default (tree, types, { search } = {}) => {
   let onSerialize = (node) =>
-    runTypeFunctionOrDefault(_.identity, types, 'onSerialize', node, {})
+    runTypeFunctionOrDefault(_.identity, types, 'onSerialize', node, {
+      extend: _.identity,
+      snapshot: _.identity,
+      initObject: _.identity,
+      log: _.identity,
```
daedalus28: We should either thread the right props through or keep this undefined. Calling `_.identity` when you think it's `extend` or `snapshot` is definitely more surprising than just having it explode as undefined.

daedalus28 (changes requested):

packages/client/src/node.js
re1ro: Addressed

packages/client/src/serialize.js
re1ro: Changed to empty object

daedalus28 (approved):

PR #81, opened 2023-05-16
client: pivot collapse fix

re1ro:
- [x] client: fixing observable trigger in collapse function

### DIFF
--- a/packages/client/src/exampleTypes/pivot.js
+++ b/packages/client/src/exampleTypes/pivot.js
@@ -113,9 +113,9 @@ export let skipResetExpansionsFields = [
 ]
 
 export default {
-  init: (node, { extend, snapshot }) => {
+  init: (node, { dispatch, extend, snapshot }) => {
     extend(node, {
-      expand(tree, type, drilldown) {
+      expand(type, drilldown) {
         drilldown = snapshot(drilldown)
         let n = snapshot(node)
         let path = n.path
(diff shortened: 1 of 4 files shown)


packages/client/src/exampleTypes/pivot.js
```diff
@@ -134,9 +134,9 @@ export default {
           ],
         })
       },
-      collapse(tree, type, drilldown) {
+      collapse(type, drilldown) {
```
stellarhoof, line 137: `expand` takes `tree` as the first argument. It's weird that `collapse` doesn't

re1ro: It's because mutate isn't exposed as an internal function. Let me see if I can address that, so they both don't need tree.

packages/client/src/index.js
```diff
@@ -252,18 +250,34 @@ export let ContextTree = _.curry(
...
+        dispatch,
+        extend,
+        snapshot,
+        initNode,
+        initObject,
+        log,
+      }),
+      tree
+    )
+    let flat = flatten(tree)
```
stellarhoof, line 267: Can't comment on this bit moving. @daedalus28 Does this look ok to you?

stellarhoof: Can't comment on this bit moving. @daedalus28 Does this look ok to you?

packages/client/src/index.js
```diff
@@ -254,16 +252,34 @@ export let ContextTree = _.curry(
...
       extend,
+      snapshot,
       types,
       initNode,
       initObject,
+      log,
     }
+    F.extendOn(
+      actionProps,
+      _.pick(
```
daedalus28, line 265: I think it's probably safe to just include all actions, but we can revisit that after dropping the flat tree

daedalus28 (approved):

PR #82, opened 2023-05-16
Deduplication Using Lowercase for Keyword Generations

positiveVibes4all:
## Summary 
Keyword generations need to be unique, regardless of the casing that is used for the words.

### DIFF
--- a/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -133,10 +133,7 @@ let ExpandableTagsQuery = ({
               />
             ))(
               _.reject(
-                _.includes(
-                  _, 
-                  keysToLower(node.context.tags)
-                ),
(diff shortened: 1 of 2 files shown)


packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -132,8 +133,8 @@ let ExpandableTagsQuery = ({
               />
             ))(
               _.reject(
-                _.includes(_, _.memoize(_.keys)(node.context.tags)),
-                _.keys(node.context.keywordGenerations)
+                _.includes(_, _.memoize(keysToLower)(node.context.tags)),
```
daedalus28: It occurs to me that the memoize makes no difference here since the function is called right away. Removing it will simplify the code and might actually slightly improve performance

daedalus28 (dismissed):

re1ro (approved):

PR #85, opened 2023-05-19
Update Adding Functionality of Keyword Generations 

positiveVibes4all:
## Summary 
Changed the source of the tags from context to the tags property so that the UI does not wait for the counts to come back to re-render. Refactored the component to its own file to help clean up the large file for Expandable tags. 

### List of updates
- Allows for the user to click as many keywords as desired while the counts load, some pages are much slower for counts to come back on and this fixes that.
- Refactor code to clean up the component file in regard to keyword generations.
- Fix height of keyword generation tag buttons

### DIFF
--- a/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -133,10 +133,7 @@ let ExpandableTagsQuery = ({
               />
             ))(
               _.reject(
-                _.includes(
-                  _, 
-                  _.memoize(keysToLower)(node.context.tags)
-                ),
+                _.includes(_, _.memoize(keysToLower)(node.context.tags)),
                 keysToLower(node.context.keywordGenerations)
               )
             )}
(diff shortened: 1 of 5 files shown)


packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -166,10 +109,15 @@ let TagsWrapper = observer(
...
     maxTags = 1000,
     hasPopover,
-    generationsCollapsed: generationsCollapse,
     enableKeywordGenerations,
+    generationsCollapsed: generationsCollapse,
     ...props
   }) => {
+    //Handle Outside Clicks for Keyword Generations display
+    let ref = React.useRef()
+    let generationsCollapsed = React.useState(true)
```
re1ro: We already have `generationsCollapsed` passed as a prop to this component, why do we need to define another one? 

If we absolutely need to define another one you need to check is there is popup open in the outside click handler like we do here

https://github.com/smartprocure/contexture/blob/349793ca2b3459ae25c5185870bf90a8904c399d/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js#L60

stellarhoof: @positiveVibes4all The entire outside click handler + `generationsCollapsed` prop + ref are not being used

positiveVibes4all: I updated the code appropriately

stellarhoof (approved):

PR #92, opened 2023-06-02
Improve performance when determining if a node has results

stellarhoof:
- Bumped futil to use `F.findNode`
- Improved performance when computing `node.hasResults`. There's no need to recurse down to every value, we just needed to find the first non-blank value.

### DIFF
--- a/packages/client/src/index.js
+++ b/packages/client/src/index.js
@@ -15,13 +15,7 @@ import actions from './actions/index.js'
 import serialize from './serialize.js'
 import traversals from './traversals.js'
 import { runTypeFunction, getTypeProp } from './types.js'
-import {
-  initNode,
-  hasContext,
-  hasValue,
-  dedupeWalk,
-  hasResults,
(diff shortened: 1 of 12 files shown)


packages/client/src/index.js
```diff
@@ -233,7 +227,7 @@ export let ContextTree = _.curry(
           if (debug && node._meta) target.metaHistory.push(node._meta)
         }
 
-        target.hasResults = hasResults(snapshot)(target)
+        target.hasResults = !!Tree.findNode(F.isNotBlank, target.context)
```
re1ro: I would still prefer it to be in a separate function. In any case you still need to `snapshot(target.context)`

re1ro (changes requested):

re1ro (approved):

packages/client/src/index.js
stellarhoof: Done

PR #109, opened 2023-06-29
Feature: pivot export

re1ro:
* [x] `export`: support for `pivot` node

### DIFF
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -30,12 +30,10 @@
   },
   "homepage": "https://github.com/smartprocure/contexture/tree/main/packages/export",
   "dependencies": {
+    "contexture-client": "^2.53.2",
     "futil": "^1.76.0",
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15"
   },
-  "packageManager": "yarn@3.3.1",
-  "peerDependencies": {
-    "contexture-client": "workspace:^"
-  }
+  "packageManager": "yarn@3.3.1"
 }
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -34,5 +34,8 @@
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15"
   },
-  "packageManager": "yarn@3.3.1"
+  "packageManager": "yarn@3.3.1",
+  "peerDependencies": {
+    "contexture-client": "workspace:^"
+  }
 }
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -30,7 +30,7 @@
   },
   "homepage": "https://github.com/smartprocure/contexture/tree/main/packages/export",
   "dependencies": {
-    "contexture-client": "^2.53.3",
+    "contexture-client": "^2.53.4",
     "futil": "^1.76.0",
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15"
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -30,7 +30,7 @@
   },
   "homepage": "https://github.com/smartprocure/contexture/tree/main/packages/export",
   "dependencies": {
-    "contexture-client": "^2.53.2",
+    "contexture-client": "^2.53.3",
     "futil": "^1.76.0",
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15"
--- a/packages/export/src/nodes/pivot.js
+++ b/packages/export/src/nodes/pivot.js
@@ -1,0 +1,130 @@
+import _ from 'lodash/fp.js'
+import F from 'futil'
+import ContextureClient from 'contexture-client';
+import { andGroup, runWith, setFilterOnly } from '../utils.js'
+import * as path from 'path'
+
+
+// TODO import from contexture-client
+let getKey = (x) => x.keyAsString || x.key
+
+// TODO import from contexture-client
+// Similar to Tree.lookup but path is a drilldown which uses keyAsString or key
+let resultsForDrilldown = (type, drilldown, results) => {
+  if (_.isEmpty(drilldown) || !results) return results
+
+  let key = _.first(drilldown)
+  let groups = _.get(type, results)
+  let match = _.find((node) => getKey(node) === key, groups)
+
+  return resultsForDrilldown(type, drilldown.slice(1), match)
+}
+
+export default async ({ service, tree, exportAllPages, ...node }) => {
+  let run = (node) => runWith(service, tree, node)
+
+  let getContextureClient = (tree, node) =>
+    ContextureClient({
+      service,
+      debounce: 0,
+    }, andGroup(setFilterOnly(tree), node))
+
+  let cardinalityNode = await run({
+    ...node,
+    expanded: {
+      columns: true,
+      rows: true,
+      skipNested: true,
+      skipValues: true,
+      cardinality: true,
+    },
+  })
+
+  let cardinalityResult =
+    _.get(
+      'context.results',
+      cardinalityNode
+    )
+
+  let getGroupingCardinality = (groupingType, result) => {
+    let getNested = _.get(groupingType)
+    let groupingCardinality = 1 // default to 1 for grand total column/row
+
+    F.walk(getNested)(
+      (groupResult, index, parents) => {
+        let groupCardinality = _.getOr(_.size(getNested(groupResult)),`${groupingType}Cardinality` , groupResult)
+        if (!groupCardinality) return
+
+        let grouping = _.get([groupingType, _.size(parents)], node)
+        if (!exportAllPages && grouping.type === 'fieldValues') {
+          groupCardinality = _.min([groupCardinality, _.getOr(10, 'size', grouping)])
+        }
+        // TODO else approximate full cardinality by using the size of the first page of each grouping
+
+        groupingCardinality = groupingCardinality + groupCardinality
+      }
+    )(result)
+
+    return groupingCardinality
+  }
+
+  let columnCardinality = getGroupingCardinality('columns', cardinalityResult)
+  let rowCardinality = getGroupingCardinality('rows', cardinalityResult)
+
+  let pivotCardinality = columnCardinality * rowCardinality * (_.size(node.values) || 1)
+
+  let pivot = {
+    node: cardinalityNode,
+    getTotalRecords() {
+      return pivotCardinality
+    },
+    async *[Symbol.asyncIterator]() {
+      let search = getContextureClient(tree, {
+        ...node,
+        expanded: {
+          columns: true,
+        },
+      })
+      await search.refresh(['root'])
+
+      let cNode = (pivot.node = search.getNode(['root-parent', node.key]))
+
+      let yieldRows = (drilldown) =>
+         async function * () {
+          let result = resultsForDrilldown('rows', drilldown, _.get('context.results', cNode))
+          let index = 0
+          let level = _.size(drilldown)
+
+          for (let row of result.rows) {
+            let path = [...drilldown, getKey(row)]
+            yield {
+              ...row,
+              path,
+              index: index++,
+              level,
+              recordCount: getGroupingCardinality('columns', row) * (_.size(cNode.values) || 1),
+            };
+
+            // if this row can have nested rows
+            if (_.size(drilldown) < _.size(cNode.rows) - 1) {
+              cNode.expand('rows', path)
+              await search.triggerUpdate()
+
+              yield * yieldRows(path)()
+
+              cNode.collapse('rows', path)
+            }
+          }
+
+          // TODO pagination if exportAllPages
+        }
+
+      yield * yieldRows([])()
+
+      let totalRow = _.get('context.results', cNode)
+
+      yield {...totalRow, path: [], index: 0, level: -1, isTotalRow: true}
+    },
+  }
+  return pivot
+}
--- a/packages/export/src/nodes/pivot.js
+++ b/packages/export/src/nodes/pivot.js
@@ -2,7 +2,6 @@ import _ from 'lodash/fp.js'
 import F from 'futil'
 import ContextureClient from 'contexture-client';
 import { andGroup, runWith, setFilterOnly } from '../utils.js'
-import * as path from 'path'
 
 
 // TODO import from contexture-client
--- a/packages/export/src/nodes/pivot.js
+++ b/packages/export/src/nodes/pivot.js
@@ -51,10 +51,7 @@ export default async ({ service, tree, exportAllPages, ...node }) => {
     let groups = _.cloneDeep(_.get(type, node))
     if (_.isEmpty(groups)) return {}
 
-    F.mergeOn(
-      _.last(groups),
-      { groupCounts: true, skip: true }
-    )
+    F.mergeOn(_.last(groups), { groupCounts: true, skip: true })
 
     if (exportAllPages) _.each(_.setOn('groupCounts', true), groups)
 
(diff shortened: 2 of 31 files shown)


packages/export/package.json
```diff
@@ -34,5 +34,8 @@
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15"
   },
-  "packageManager": "yarn@3.3.1"
+  "packageManager": "yarn@3.3.1",
+  "peerDependencies": {
```
stellarhoof: This is a regular dep, not a peer dep. Peer deps won't be installed automatically and you need this to run exports so it's not optional.

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,139 @@
...
+
+  let pivotCardinality =
+    columnCardinality * rowCardinality * (_.size(node.values) || 1)
+
+  let pivot = {
+    node: cardinalityNode,
+    getTotalRecords() {
+      return pivotCardinality
+    },
+    async *[Symbol.asyncIterator]() {
```
stellarhoof, line 101: I know you were copying other code but this is probably easier to read:

```javascript
async function* pivot() {
   ....
}

pivot.node = ...

pivot.getTotalRecords = ...

return pivot
```

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,139 @@
...
+        expanded: {
+          columns: true,
+        },
+      })
+      await search.refresh(['root'])
+
+      let cNode = (pivot.node = search.getNode(['root-parent', node.key]))
+
+      let yieldRows = (drilldown) =>
+        async function* () {
```
stellarhoof: No need to wrap the inner fn. You can do

```javascript
async function* yieldRows(drilldown) {
     ...
     yield* yieldRrows(...)
     ...
}
```

stellarhoof:
@re1ro Can you add a description on how to test this PR? I may not be able to review this PR in depth but we can all test it out.

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -292,20 +292,48 @@ let createPivotScope = (node, schema, getStats) => {
...
+              )([
+                {
+                  type: 'cardinality',
+                  field: group.field,
+                  precision_threshold: 100, // less load on ES
+                },
+              ]),
+            },
+          })
+        }
```
daedalus28: We should not allow type-specific details to leak into pivot.js itself

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,139 @@
...
+// Similar to Tree.lookup but path is a drilldown which uses keyAsString or key
+let resultsForDrilldown = (type, drilldown, results) => {
+  if (_.isEmpty(drilldown) || !results) return results
+
+  let key = _.first(drilldown)
+  let groups = _.get(type, results)
+  let match = _.find((node) => getKey(node) === key, groups)
+
+  return resultsForDrilldown(type, drilldown.slice(1), match)
+}
```
daedalus28: This code is also in contexture-elasticsearch and has no unit tests there either. I swapped it locally for a version that used `Tree.lookup`, but without any tests I couldn't easily tell if it was equivalent.

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+  let cardinalityNode = await run({
+    ...node,
+    expanded: {
+      columns: true,
+      rows: true,
+      skipNested: true,
+      skipValues: true,
+      cardinality: true,
```
daedalus28: I'm not sure it makes sense to extend the pivot API itself with these features - it might just make more sense to implement a simple transform here that just replaces `values` for `cardinality`, and drop the innermost groupings instead of a `skipNested` api flag.

The idea is to keep this part as a very straightforward, clearly unit testable transform of json -> json.

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,139 @@
...
+              level,
+              recordCount:
+                getGroupingCardinality('columns', row) *
+                (_.size(cNode.values) || 1),
+            }
+
+            // if this row can have nested rows
+            if (_.size(drilldown) < _.size(cNode.rows) - 1) {
+              cNode.expand('rows', path)
+              await search.triggerUpdate()
```
daedalus28: `expand` triggers a mutate so you should be able to await it directly without needing `triggerUpdate` - just return here: https://github.com/smartprocure/contexture/blob/8ab95ebdc4fbc0aca2b5bbfba11224e7ff047745/packages/client/src/exampleTypes/pivot.js#L126

In general, `triggerUpdate` is usually a code smell and not what you want because contexture is reactive.

packages/export/src/nodes/pivot.js
daedalus28: Let's stay consistent with the rest other ones for now - we're probably more likely to change the structure to `{ getTotalRecords, async* run }` or something like that than to dig deeper in this structure.

packages/export/package.json
```diff
@@ -34,5 +34,8 @@
...
+    "contexture-client": "workspace:^"
```
daedalus28: I don't think we should be shipping with `workspace` in our dependencies 🤔

yarn.lock
```diff
@@ -6479,6 +6479,8 @@ __metadata:
     futil: ^1.76.0
     lodash: ^4.17.21
     minimal-csv-formatter: ^1.0.15
+  peerDependencies:
+    contexture-client: "workspace:^"
```
daedalus28: I don't think we should be shipping with `workspace` in our dependencies 🤔

daedalus28 (changes requested):
This is just a first review pass - I expect we'll have follow on comments after these changes since they'll make it clearer to see what's going on.

yarn.lock
re1ro: Having `workspace` allows you to use the local version when developing instead of pulling a package from NPM. 
We could do it for all the cross-dependencies.
Supposedly changeset should replace workspace with correct version on release
https://github.com/changesets/changesets/issues/432

packages/export/package.json
re1ro: Having workspace allows you to use the local version when developing instead of pulling a package from NPM.
We could do it for all the cross-dependencies.
Supposedly changeset should replace workspace with correct version on release
https://github.com/changesets/changesets/issues/432

stellarhoof: `workspace` gets replaced by the correct version before publishing

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -292,20 +292,44 @@ let createPivotScope = (node, schema, getStats) => {
...
+          parent = _.merge(parent, {
+            aggs: {
+              [`${groupingType}Cardinality`]: {
+                cardinality: {
+                  field: getField(schema, group.field),
+                  precision_threshold: 100, // less load on ES
+                },
+              },
+            },
+          })
```
daedalus28: Let's move this logic into type-specific `getGroupCount` methods since `cardinality` is `fieldValues` specific

The default implementation just relies on bucket count. `dateHistogram` could do min/max date with math, etc. `numberRanges` would return the number of ranges from the node, `numberInterval` would min/max (e.g. `(max-min)/interval`),  `tagsQuery` returns count of tags,  etc

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -292,20 +292,44 @@ let createPivotScope = (node, schema, getStats) => {
...
-        let parent = await build(
-          group,
-          await children,
-          groupingType,
-          schema,
-          getStats,
-          drilldownKey
-        )
+        let withGroupingCardinality =
+          _.get('expanded.cardinality', node) &&
```
daedalus28: this flag becomes `groupCounts`

daedalus28: Changes:

- Move `groupCounts: true | false` to a per grouping config value
- Add `skip: true | false` per grouping

This will enable UI features to show paging info within the UI

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,156 @@
...
+        [type]: true,
+        skipValues: true,
+        cardinality: hasFieldValuesLast || !!exportAllPages,
+      },
+    })
+
+    return _.get('context.results', cardinalityNode)
+  }
+
+  let getGroupingSize = (groupingType, cardinalityResult) => {
```
daedalus28: let's add tests for this

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,156 @@
...
+    let getNested = _.get(groupingType)
+    let groupingCardinality = 1 // starting with 1 for the total column/row
+
+    F.walk(getNested)((groupResult, index, parents) => {
+      let groupCardinality = _.getOr(
+        _.size(getNested(groupResult)),
+        `${groupingType}Cardinality`,
+        groupResult
+      )
```
daedalus28: Now that we'll have `groupCounts` natively in `provider-elasticsearch`, you can make it default to the bucket size on the server when `groupCounts` are true. Moving this logic also enables support for this in an eventual UI feature without having to duplicate the thinking.

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,156 @@
...
+      if (!groupCardinality) return
+
+      let grouping = _.get([groupingType, _.size(parents)], node)
+      if (!exportAllPages && grouping.type === 'fieldValues') {
+        groupCardinality = _.min([
+          groupCardinality,
+          _.getOr(10, 'size', grouping),
+        ])
```
daedalus28: Eventually, this will also move out since we'll want paging info done in type definitions on the client (e.g. you can check a method like `hasMore` in the future)

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,156 @@
...
+        `${groupingType}Cardinality`,
+        groupResult
+      )
+      if (!groupCardinality) return
+
+      let grouping = _.get([groupingType, _.size(parents)], node)
+      if (!exportAllPages && grouping.type === 'fieldValues') {
+        groupCardinality = _.min([
+          groupCardinality,
+          _.getOr(10, 'size', grouping),
```
daedalus28: I'm not sure you need this

daedalus28: I'm not sure you need this. If you do, we should add a comment explaining when/why

daedalus28: You can also factor out the type specific stuff so it's clear like this:
```js
let getPageSize = (grouping) => {
  let fns = {
    fieldValues: _.getOr(10, "size"),
  }
  return (fns[grouping.type] || _.noop)(grouping)
}
if (!exportAllPages)
  groupCardinality = _.min([groupCardinality, getPageSize(grouping)])
```

and then move to the client later

daedalus28 (changes requested):

packages/export/src/nodes/pivot.js
re1ro: Imported from client for now

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,176 @@
...
+      },
+      andGroup(setFilterOnly(tree), node)
+    )
+
+  let getGroupingResult = async (type) => {
+    let groups = _.cloneDeep(_.get(type, node))
+    let lastGroup = _.last(groups)
+    if (_.isEmpty(groups)) return {}
+
+    if (lastGroup.type === 'fieldValues')
```
daedalus28: Do you still need this check? The elasticsearch provider handles other types now.

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,176 @@
...
+  let pageSizeGetters = {
+    fieldValues: _.getOr(10, 'size'),
+  }
+  return _.getOr(_.noop, grouping.type, pageSizeGetters)(grouping)
+}
+
+export let getGroupingSize = (
+  node,
+  groupingType,
+  cardinalityResult,
```
daedalus28: There are lots of references here to `cardinality`, which is only correct for fieldValues. Now that the elasticsearch provider handles other cases, we should match the terminology

packages/export/src/nodes/pivot.js
```diff
@@ -0,0 +1,176 @@
...
+  exportAllPages
+) => {
+  let getNested = _.get(groupingType)
+  let groupingCardinality = 1 // starting with 1 for the total column/row
+
+  F.walk(getNested)((groupResult, index, parents) => {
```
daedalus28: This would be cleaner as a reduce. Here's some quick (not tested) ⛳ 

```js
export let getGroupingSize = (node, groupType, countResult, exportAllPages) =>
  F.reduceTree(_.get(groupType))(
    (result, groupResult, index, parents) => {
      let childSize = _.size(_.get(groupType, groupResult))
      let groupSize = _.getOr(childSize, `${groupType}GroupCount`, groupResult)

      if (!exportAllPages) {
        let grouping = _.get([groupType, _.size(parents)], node)
        groupSize = _.min([groupCount, getPageSize(grouping)])
      }

      // TODO else approximate full cardinality by using the average size of the first page of each grouping
      return result + groupSize
    },
    1, // starting with 1 for the total column/row
    countResult,
  )
```

daedalus28: It might even be clearer as a `_.sum` on `mapTree`

daedalus28 (changes requested):

packages/export/src/nodes/pivot.js
re1ro: We should keep it for now, because setting explicit `skip` for non `fieldValues` will skip the sub-aggregation and `groupCount` retrieval. Until we write the script aggregations for other types

packages/export/src/nodes/pivot.js
re1ro: I think you meant `treeToArrayBy` istead of `mapTree`, but it's not really cleaner, so I will go with the reduce
```js
export let getGroupingSize = (node, groupType, countResult, exportAllPages) =>
  _.flow(
    F.treeToArrayBy(_.get(groupType))(
      (groupResult, index, parents) => {
        let childSize = _.size(_.getOr(0, groupType, groupResult))
        let groupSize = _.getOr(childSize, `${groupType}GroupCount`, groupResult)

        if (!groupSize) return 0

        if (!exportAllPages) {
          let grouping = _.get([groupType, _.size(parents)], node)
          groupSize = _.min([groupSize, getPageSize(grouping)])
        }

        // TODO else approximate full cardinality by using the average size of the first page of each grouping
        return groupSize
      },
    ),
    _.sum,
    _.add(1) // add 1 for the grand total row/column
  )(countResult)
  ```

File: packages/export/src/nodes/pivot.test.js [hunk 19]
```diff
@@ -0,0 +1,260 @@
...
+        rows: undefined,
+      }),
+      pivotResult
+    )
+    expectedRecords.push({ ...expectedRecords.shift(), isTotalRow: true })
+
+    let records = []
+    for await (const r of strategy) records.push(r)
+
+    let exportedRecordCount = _.reduce(
```
daedalus28: This is just a _.sumBy

packages/export/src/nodes/pivot.js
daedalus28: If we don't do that now, I'm not sure we ever will. Alternatively, we could support checking whether the type implements getGroupCount or not (which could even just live in the elasticsearch provider as better skip behavior - either as the only behavior for skip or a new type skip mode)

daedalus28 (dismissed):

stellarhoof (approved):

PR #124, opened 2023-08-09
ES pivot: `groupCounts` and `skip` configuration

re1ro:
* [x] ElasticSearch pivot: adding `groupCounts` and `skip` per group/value config. `groupCounts` flag adds subgroups counter to the parent level. `skip` flag skips the groping level in aggregation

### DIFF
--- a/.changeset/strange-plants-pull.md
+++ b/.changeset/strange-plants-pull.md
@@ -2,4 +2,4 @@
 'contexture-elasticsearch': minor
 ---
 
-Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to the parent level. `skip` flag skips the groping level in aggregation 
+Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to the parent level. `skip` flag skips the groping level in aggregation
--- a/.changeset/strange-plants-pull.md
+++ b/.changeset/strange-plants-pull.md
@@ -2,4 +2,4 @@
 'contexture-elasticsearch': minor
 ---
 
-Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to parent level. `skip` flag skips the groping level in aggregation
+Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to the parent level. `skip` flag skips the groping level in aggregation 
--- a/.changeset/strange-plants-pull.md
+++ b/.changeset/strange-plants-pull.md
@@ -2,4 +2,4 @@
 'contexture-elasticsearch': minor
 ---
 
-Pivot: configuring `groupCounts` and `skip` per group/value level
+Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to parent level. `skip` flag skips the groping level in aggregation 
--- a/.changeset/strange-plants-pull.md
+++ b/.changeset/strange-plants-pull.md
@@ -2,4 +2,4 @@
 'contexture-elasticsearch': minor
 ---
 
-Pivot: configuring `groupCounts` and `skip` per group/value level. `groupCounts` flag adds subgroups counter to parent level. `skip` flag skips the groping level in aggregation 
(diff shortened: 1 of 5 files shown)


packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -289,23 +289,38 @@ let createPivotScope = (node, schema, getStats) => {
           // The API of `{sort: {field, direction}}` is respected by fieldValues and can be added to others
           group.sort = { field: sortField, direction: sort.direction }
         }
-        let build = lookupTypeProp(_.identity, 'buildGroupQuery', group.type)
-        //Remove anything that needs to be hoisted before composing further
-        // We are iterating through the groups reversed so we need to subtract instead of add to get the right index
-        let reversedLookupIndex = groups.length - index - 1
+        let build = lookupTypeProp(
+          _.rearg([1], _.identity),
```
stellarhoof: An inline fn would be less confusing

.changeset/strange-plants-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'contexture-elasticsearch': minor
+---
+
+Pivot: configuring `groupCounts` and `skip` per group/value level
```
stellarhoof: Can you add a better description? What does `groupCounts` do? What does `skip` do?

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -362,7 +377,10 @@ let createPivotScope = (node, schema, getStats) => {
       groups: rows,
     })
 
-    let statsAggs = { aggs: getAggsForValues(node.values) }
+    let aggValues = _.reject('skip', node.values)
+    let statsAggs = _.isEmpty(aggValues)
+      ? {}
```
daedalus28, line 382: Do we need special handling for this? We didn't handle this before if node.values was empty

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
daedalus28: I'm not sure we should do this at all. This change fixes a bug to return `children` instead of incorrectly returning `group` when pivot groupings don't specify `buildGroupQuery`, but when would that be desirable? Should the default case throw an error?

.changeset/strange-plants-pull.md
daedalus28: Agreed - this should also be in the readme.

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
re1ro: Agreed, replacing it with a throw function instead

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
re1ro: We didn't it just produces a strange query with {aggs: {}} nested in it. 
I've added this case handling to produce cleaner queries

packages/provider-elasticsearch/src/example-types/metricGroups/pivot.js
```diff
@@ -289,23 +289,38 @@ let createPivotScope = (node, schema, getStats) => {
           // The API of `{sort: {field, direction}}` is respected by fieldValues and can be added to others
           group.sort = { field: sortField, direction: sort.direction }
         }
-        let build = lookupTypeProp(_.identity, 'buildGroupQuery', group.type)
-        //Remove anything that needs to be hoisted before composing further
-        // We are iterating through the groups reversed so we need to subtract instead of add to get the right index
-        let reversedLookupIndex = groups.length - index - 1
+        let build = lookupTypeProp(
+          () => F.throws(Error('Unknown grouping type')),
```
daedalus28: Let's put the type name in the error message

daedalus28 (changes requested):

daedalus28 (approved):

daedalus28 (approved):

PR #139, opened 2023-09-26
Misc advanced search bugfixes

stellarhoof:
Generally improve styling across the board and fix drop targets

Before

Screenshot 2023-09-27 at 12 57 07 AM

After

Screenshot 2023-09-27 at 12 59 20 AM


daedalus28 (approved):

stellarhoof:
Follow up fixes:

- Comment about the operator layout logic
- Allow drop target under rule to work on margins
- Single rule NOT group are missing operator tag

PR #146, opened 2023-10-02
Refactor: break down getAdditionalFields

stellarhoof:
Break down another big function into smaller functions.

Also `anyRegexesMatch` was broken so this PR fixes it and includes a couple unit tests for it.

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/highlighting.js
+++ b/packages/provider-elasticsearch/src/example-types/results/highlighting.js
@@ -1,13 +1,8 @@
 import F from 'futil'
 import _ from 'lodash/fp.js'
 
-function regexify(criteria) {
-  return new RegExp(criteria)
-}
-
-function anyRegexesMatch(regexes, criteria) {
-  return _.some(_.invokeMap(_.map(regexify, regexes), 'test', criteria))
-}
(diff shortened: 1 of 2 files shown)


packages/provider-elasticsearch/src/example-types/results/highlighting.js
```diff
@@ -1,13 +1,8 @@
...
 
-function regexify(criteria) {
-  return new RegExp(criteria)
-}
-
-function anyRegexesMatch(regexes, criteria) {
-  return _.some(_.invokeMap(_.map(regexify, regexes), 'test', criteria))
-}
+export let anyRegexesMatch = (regexes, criteria) =>
+  !!_.find((str) => new RegExp(str).test(criteria), regexes)
```
positiveVibes4all: I think it might be better for str to be regex or pattern or something like that for readability purposes. You do not have to but it just seems like you made it way more readable already and this would help in that regard slightly while doing this.

positiveVibes4all (approved):
Left one comment about the refactor that seems to make sense.

packages/provider-elasticsearch/src/example-types/results/highlighting.js
stellarhoof: Done

PR #147, opened 2023-10-03
Add highlighting documentation

stellarhoof:
- Add typescript types for schema/node highlighting properties
- Improve performance and robustness in the code that removes/tests for highlighting tags


positiveVibes4all (approved):
Really like the typescript file for documentation purposes!

stellarhoof:
@positiveVibes4all Yeah it's something we can use today and possibly even export the types once coverage is good enough

PR #152, opened 2023-10-16
Merge results node highlighting settings into schema highlighting settings

stellarhoof:
We added the ability to override schema highlighting configuration in the results node in https://github.com/smartprocure/contexture/pull/64. However, it is not convenient as the node overrides replace the schema highlighting config instead of merging into them, leading users to have to repeat the schema highlighting config everywhere they need to override it.

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/highlighting.js
+++ b/packages/provider-elasticsearch/src/example-types/results/highlighting.js
@@ -201,22 +201,26 @@ export let highlightResults = ({
   return { additionalFields }
 }
 
+const mergeReplacingArrays = _.mergeWith((target, src) => {
+  if (_.isArray(src)) return src
+})
+
 export let getHighlightSettings = (schema, node) => {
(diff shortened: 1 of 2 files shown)


packages/provider-elasticsearch/src/example-types/results/highlighting.js
```diff
@@ -253,7 +257,20 @@ export let getHighlightSettings = (schema, node) => {
             _.pick(_.concat(node.include, schemaInlineAliases), filtered)
     )(schemaHighlight)
 
-    nodeHighlight = _.merge(
+    // Properties we support as part of the highlighting configuration that
+    // elastic does not have knowledge of.
+    let nonElasticProperties = [
```
daedalus28, line 262: This is fine for now, but let’s made sure we have something tracking the removal of these properties in favor of simpler options (eg using include instead of inline)

daedalus28 (approved):

PR #154, opened 2023-10-24
Update Highlighting Algorithm

positiveVibes4all:
## Summary 
This PR updates the highlighting algorithm to not rely on setting require field match to false. This fixes instances where filters that did not originate from copy_to configurations do not highlight all fields and instead just the field that they are for. This PR also merges highlighting when there is an instance of a field to highlight and one of its related subfields as well.


packages/provider-elasticsearch/src/example-types/results/highlighting.test.js
```diff
@@ -394,3 +396,55 @@ describe('containsHighlightTagRegex()', () => {
     expect(regex.test(text)).toEqual(false)
   })
 })
+
+describe('Highlight field aggregation', () => {
```
stellarhoof: We should name the describe block after the fn is being tested. I added the article that I mostly follow when writing unit tests in notion: https://www.notion.so/Best-Practices-0d931662db624a77af8dd09cc41e01bf

packages/provider-elasticsearch/src/example-types/results/index.js
```diff
@@ -61,6 +61,7 @@ export default {
             nodeHighlight, // The result node's highlight configuration
             hit, // The ES result
             include: resultColumns, // The columns to return
+            node, // The subfields to return
```
stellarhoof: Comment's wrong

packages/provider-elasticsearch/src/schema.js
```diff
@@ -29,17 +29,22 @@ let fromEsIndexMapping = (mapping) => {
...
             flatten,
-            F.mapValuesIndexed(({ type, fields }, field) => ({
+            F.mapValuesIndexed(({ type, fields, copy_to }, field) => ({
               field,
               label: _.startCase(field),
               elasticsearch: F.compactObject({
                 dataType: type,
                 // Find the child notAnalyzedField to set up facet autocomplete vs word
                 notAnalyzedField: _.findKey({ type: 'keyword' }, fields),
+                ...(copy_to && { copy_to }),
```
stellarhoof: No need to spread it, just add `copy_to` inline

packages/provider-elasticsearch/src/utils/highlightUtil.js
stellarhoof, line 1: You probably added this util at the top because of Sam's comment on making it generic but it's really only relevant to the highlighting code so we should move it there.

packages/provider-elasticsearch/src/utils/highlightUtil.test.js
stellarhoof, line 1: Can we use more compact highlight tags such as `<em></em>`? It makes it easier to follow the code.

packages/provider-elasticsearch/src/utils/highlightUtil.test.js
```diff
@@ -0,0 +1,58 @@
...
+      'foo.bar': [
+        '<b class="search-highlight">foo</b> this <b class="search-highlight">handle</b>',
+      ],
+      foo: [
+        '<b class="search-highlight">foo</b> this <b class="search-highlight">bar</b>',
+      ],
+      'foo.car.bar': ['<b class="search-highlight">foo</b> this is not merged'],
+    })
+  })
+  it('should combine subfield highlights with field highlights1', () => {
```
stellarhoof: This function serves the same purpose as the one above. Also the field values are noisy so it's difficult to detect where the highlighted substrings are.

packages/provider-elasticsearch/src/utils/highlightUtil.test.js
```diff
@@ -0,0 +1,58 @@
+import { mergeHitHighlights } from './highlightUtil.js'
+
+describe('Highlight field aggregation', () => {
```
stellarhoof: There are some important test cases we can add here:
1. Test when a highlighted string is contained in another
2. Test when a highlighted string overlaps with another

packages/provider-elasticsearch/src/utils/highlightUtil.js
```diff
@@ -0,0 +1,68 @@
...
+let removeHighlightTags = (nodeHighlight, highlightString) =>
+  _.replace(findHighlightTagsRegEx(nodeHighlight), '', highlightString)
+
+let getFieldFromSubField = _.flow(
+  F.dotEncoder.decode,
+  _.dropRight(1),
+  F.dotEncoder.encode
+)
+
+export let mergeHitHighlights = (nodeHighlight, fields, hitHighlights) => {
```
stellarhoof: This generic fn should only take strings only, e.g. `mergeHitHighlights(hi1, hi2, ...)`

packages/provider-elasticsearch/src/example-types/results/highlightUtil.js
```diff
@@ -0,0 +1,71 @@
...
+let getHighlightMatches = _.curry((nodeHighlight, highlight) =>
+  highlight.match(highlightContentRegEx(nodeHighlight))
+)
+
+let getWordListMatchesRegEx = (words) => new RegExp(words.join('|'), 'gi')
+
+let highlightWith = ({ pre_tags, post_tags }) =>
+  F.highlight(pre_tags[0], post_tags[0])
+
+let findHighlightTagsRegEx = ({ pre_tags, post_tags }) =>
```
stellarhoof, line 16: This is the same regex as https://github.com/smartprocure/contexture/blob/main/packages/provider-elasticsearch/src/example-types/results/highlighting.js#L7

packages/provider-elasticsearch/src/example-types/results/highlightUtil.js
```diff
@@ -0,0 +1,71 @@
+import _ from 'lodash/fp.js'
+import F from 'futil'
+
+let highlightContentRegEx = ({ pre_tags, post_tags }) =>
+  new RegExp(`(?<=${pre_tags[0]})(.*?)(?=${post_tags[0]})`, 'g')
```
stellarhoof, line 5: Lookaheads/lookbehinds are generally expensive. You can avoid them by using `matchAll` instead of `match`. Ex:

```javascript
const regex = new RegExp('<em>(.*?)</em>', 'g')
const str = 'a <em>b</em> c <em>d e</em> f'
const matches = Array.from(str.matchAll(regex)).map(_.nth(1))
```

Also, avoid compiling (instantiating) regexes multiple times if it can be avoided. For example `new RegExp('<em>(.*?)</em>', 'g')` can be compiled once and used many times inside of a loop.

packages/provider-elasticsearch/src/example-types/results/highlightUtil.js
```diff
@@ -0,0 +1,71 @@
...
+        : undefined
+      if (field && hitHighlights[field] && hitHighlights[subField]) {
+        let strippedField = removeHighlightTags(
+          { pre_tags, post_tags },
+          hitHighlights[field]
+        )
+        let highlightPostings = F.flowMap(
+          _.head,
+          getHighlightMatches({ pre_tags, post_tags }),
+          getWordListMatchesRegEx,
```
stellarhoof, line 48: This regex shouldn't be case insensitive

packages/provider-elasticsearch/src/example-types/results/highlightUtil.test.js
```diff
@@ -0,0 +1,55 @@
+import { mergeHitHighlights } from './highlightUtil.js'
+
+let nodeHighlight = {
+  pre_tags: ['<em>'],
+  post_tags: ['</em>'],
+}
+
+describe('mergeHitHighlights()', () => {
```
stellarhoof, line 8: Another test case we should add is when there are identical substrings being highlighted more than once. Ex: `'<em>lazy</em> for was really <em>lazy</em>`

packages/provider-elasticsearch/src/example-types/results/highlightUtil.js
```diff
@@ -0,0 +1,71 @@
...
+let removeHighlightTags = (nodeHighlight, highlightString) =>
+  _.replace(findHighlightTagsRegEx(nodeHighlight), '', highlightString)
+
+let getFieldFromSubField = _.flow(
+  F.dotEncoder.decode,
+  _.dropRight(1),
+  F.dotEncoder.encode
+)
+
+export let mergeHitHighlights = (
```
stellarhoof, line 28: Now that I've read this fn I think it's easier if we do a single flow where we handle all the postings upfront. Ex:

```javascript
_.flow(
  _.flatMap(getHighlightedWords, getPostings),
  F.mergeRanges,
  highlightRanges,
)
```

PR #156, opened 2023-11-01
Revamp highlighting

stellarhoof:
Completely overhaul the highlighting API in the elasticsearch `results` example type. Check [the documentation](https://github.com/smartprocure/contexture/blob/e12dfca069b0dad4f1bd730462a5b60e14a7ff44/packages/provider-elasticsearch/src/example-types/results/README.md) for details.


packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
```diff
@@ -0,0 +1,33 @@
...
+  if (config?.filterSourceArrays) {
+    for (const path of getArrayFieldsPaths(schema)) {
+      if (!_.has(path, highlights)) {
+        F.setOn(path, [], highlights)
+      }
+    }
+  }
+
+  // Mutate source only for performance reasons
+  _.convert({ immutable: false }).mergeWith(
```
daedalus28: This seems related to `F.mergeAllArrays` and `F.mergeOverAllArrays`. We might want to factor out a more generic futil candidate even if we don't add it to futil.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,108 @@
+import _ from 'lodash/fp.js'
+import F from 'futil'
+import { CartesianProduct } from 'js-combinatorics'
+
+/**
+ * Return names for all fields groups and their sub-fields in `schema`.
+ */
+const getFieldsGroupsNames = _.memoize((schema) => {
```
daedalus28: We should export these methods and add tests for them

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,108 @@
...
+}, _.get('elasticsearch.index'))
+
+/**
+ * Return mappings for all sub-fields that can be highlighted in `mapping`.
+ */
+const getSubFieldsMappings = (schema, mapping) =>
+  F.reduceIndexed(
+    (acc, sfMapping, sfName) => {
+      if (schema.elasticsearch.subFields?.[sfName]?.highlight) {
+        acc[sfName] = {
```
daedalus28: I think this could be coded more cleanly with mapValues - but let's add some tests to this function either way. With tests, it'd be easier to have a drop in refactoring suggestion.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,108 @@
...
+const getFieldsGroupsNames = _.memoize((schema) => {
+  const subFields = _.keys(
+    _.pickBy('highlight', schema.elasticsearch?.subFields)
+  )
+
+  const fields = _.flatMap((field) => {
+    const copy_to = field.elasticsearch?.copy_to
+    if (!_.isEmpty(copy_to)) {
+      const product = new CartesianProduct(copy_to, subFields)
+      return [...copy_to, ...Array.from(product).map(_.join('.'))]
```
daedalus28: Pretty sure `Array.from(product).map(_.join('.'))` can be reduced to `_.map(_.join('.'), product)`

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,108 @@
...
+ * Return names for all fields groups and their sub-fields in `schema`.
+ */
+const getFieldsGroupsNames = _.memoize((schema) => {
+  const subFields = _.keys(
+    _.pickBy('highlight', schema.elasticsearch?.subFields)
+  )
+
+  const fields = _.flatMap((field) => {
+    const copy_to = field.elasticsearch?.copy_to
+    if (!_.isEmpty(copy_to)) {
```
daedalus28: an early return would be simpler here since you could avoid nesting - `if (_.isEmpty(copy_to)) return copy_to`

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Returns an array of [start, end] ranges that correspond to substrings
+ * enclosed in pre/post tags. The ranges correspond to the plain string without
+ * tags. For example given the tags `{ pre: '<em>', post: '</em>' }`, this
+ * function will return [[2, 5], [6, 9]] for the string
+ *
+ * `A <em>red</em> <em>car</em>`
+ */
+const getHighlightRanges = (pre, post, str) => {
```
daedalus28: Is there a reason we couldn't use `F.postings` or `F.postingsForWords`?

packages/provider-elasticsearch/src/utils/futil.js
```diff
@@ -99,3 +99,21 @@ export let renameOn = (from, to, obj) => {
 // Async version of compactMap (and indexed)
 export let compactMapAsync = async (...args) =>
   _.compact(await Promise.all(F.mapIndexed(...args)))
+
+// Similar to F.getOrReturn but for _.set
+export const setOrReturn = (path, val, obj) =>
+  path ? _.set(path, val, obj) : val
+
+// _.groupBy but also passing the current key
+export const groupByIndexed = _.curry((it, coll) =>
```
daedalus28, line 106: We should be able to call convert on lodash instead of reimplementing `groupBy`. I think we also expose `F.indexed` so you could do `F.indexed.groupBy`

packages/provider-elasticsearch/src/example-types/results/type.d.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Typings for the result example type.
+ */
+
+interface HighlightConfig {
+  /**
+   * Whether to send highlighting configuration to elastic and merge
+   * highlighting results onto source. Defaults to `false`.
+   */
+  enable?: boolean
```
daedalus28: I think the default behavior should be to highlight, so we should invert this to a `disable` flag

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
stellarhoof: This custom merging fn removes elements from arrays based on some flag and lodash/futil merges do not prune the result. Also this fn requires arrays in the right-hand side object to have a very specific structure (they may contain `undefined` to fill in for non-highlighted items).

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: All the functions here are well tested indirectly through the exported function in this file. I did this to avoid testing implementation details since the semantics of the exported fn is not that complicated to test.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: `product` is a custom iterator returned by combinatorics-js and map doesn't know how to consume it.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: Yeah this is a flow of `pickBy` + `mapValues`.

Same comment as above about tests.

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
stellarhoof: `postings` is pretty different since it returns postings for the entire regex match instead of just the capture

```javascript
const str = 'The <em>lazy</em> fox'

 // -> [ [ 4, 17 ] ]
F.postings(new RegExp('<em>(.*?)</em>', 'g'), str)

 // -> [ [ 4, 8 ] ]
getHighlightedRanges('<em>', '</em>', str)
```

Also this fn pretends that tags are 0 width which postings doesn't do. Notice the difference in the end positions above (17 vs 8).

packages/provider-elasticsearch/src/example-types/results/type.d.ts
stellarhoof: Highlighting is a perf hit so I don't think we should make it opt-out. Also highlighting is optional, I think we even have more searches that don't need highlighting in the product.

packages/provider-elasticsearch/src/utils/futil.js
stellarhoof: The normal (non-fp) `groupBy` does not pass the index so that won't work

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: Got rid of the reduce in favor of pickBy+mapValues

.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": ["eslint:recommended", "plugin:import/recommended"],
-  "ignorePatterns": ["dist", "node_modules"],
+  "ignorePatterns": ["dist", "node_modules", "*.ts"],
```
daveyarwood, line 3: Is this intentional?

stellarhoof: Yes, we don't have typescript enabled and eslint errors out on ts files (I introduced a `d.ts` file for documentation purposes).

daveyarwood: Got it, makes sense 👍

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
```diff
@@ -0,0 +1,33 @@
...
+import F from 'futil'
+import { getArrayFieldsPaths } from './util.js'
+
+export const mergeHighlightsOnSource = (schema, config, source, highlights) => {
+  // Account for an edge case where source arrays should only contain
+  // highlighted items but there are no highlights in the results.
+  if (config?.filterSourceArrays) {
+    for (const path of getArrayFieldsPaths(schema)) {
+      if (!_.has(path, highlights)) {
+        F.setOn(path, [], highlights)
```
positiveVibes4all: This seems like a reimplementation of defaults, it might be more intuitive to have something like this instead
```
F.defaultsOn(
   F.keysToObject(_.stubArray, getArrayFieldsPaths(schema)), 
   highlights
)
```

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
```diff
@@ -0,0 +1,33 @@
...
+  _.convert({ immutable: false }).mergeWith(
+    (src, hi) => {
+      if (_.isArray(src) && config?.filterSourceArrays) {
+        return F.reduceIndexed(
+          (acc, v, i) =>
+            _.isUndefined(v)
+              ? acc
+              : F.push(_.isPlainObject(v) ? _.merge(src[i], v) : v, acc),
+          [],
+          hi
```
positiveVibes4all: This is not necessary but I think being more verbose here and just naming 'hi' as 'highlights' and naming 'v' as 'highlight' would help for readability.

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.test.js
```diff
@@ -0,0 +1,201 @@
...
+      state: 'California',
+      city: 'San Francisco',
+      book: {
+        title: 'Don Quixote',
+        blob: 'The quick brown fox jumped over the lazy dog',
+        authors: ['John Snow', 'Neo'],
+      },
+    }
+    const highlights = {
+      city: '<em>San</em> Francisco',
```
positiveVibes4all: Highlights always come over in the with an array or am I missing some case here that I am not aware of?
Highlight

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.test.js
```diff
@@ -0,0 +1,201 @@
...
+      book: {
+        blob: ['<em>The</em> quick', 'jumped <em>over</em>'],
```
positiveVibes4all: What highlights come over like this? I have not see them come over nested I always see them in a flattened path.

packages/provider-elasticsearch/src/example-types/results/type.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Whether to remove non-highlighted items in source arrays. Defaults to
+   * `false`.
+   */
+  filterSourceArrays?: boolean
```
positiveVibes4all: This name feels a little confusing to me, some ideas I thought about that seem like a better fit: 

- onlyArrayHighlights/isOnlyArrayHighlights
- onlySourceArrayMatches/isOnlySourceArrayMatches

I am not saying that you should change it, but considering something else like this would help when reading the code IMO.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,100 @@
...
+  const getHighlightQuery = (mapping, name) => {
+    const toReplace = _.intersection(queryFieldsGroups, mapping.copy_to)
+    if (!_.isEmpty(toReplace)) {
+      const regexp = new RegExp(_.join('|', toReplace), 'g')
+      return JSON.parse(_.replace(regexp, name, querystr))
+    }
+  }
+
+  const mappingToHighlightConfig = (mapping, name) => {
+    const isBlob = mapping.meta?.subType === 'blob'
```
positiveVibes4all: Maybe we can put the fragment size and number of fragments in the meta data here so that it is configurable based on the mappings with the values here as sane defaults still?

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
positiveVibes4all: When I was doing this prior, I thought there was a good opportunity here to update postings for words wrapped in html based on an HTML tag or something like this, but this use case is much different than what those two functions do out of the box today.

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
stellarhoof: Lodash's `defaults` does not handle nested paths: `_.defaults({ 'a.b': 1 }, { a: {} }) => { a: {}, 'a.b': 1 }`

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.test.js
stellarhoof: These highlights are not the ones that come back from elastic; they have been transformed by https://github.com/smartprocure/contexture/blob/feature/revamp-highlighting/packages/provider-elasticsearch/src/example-types/results/highlighting/response.js#L20. It's all in the documentation.

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.test.js
stellarhoof: Same comment as above.

packages/provider-elasticsearch/src/example-types/results/type.d.ts
stellarhoof: `removeUnmatchedArrayItems`?

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: Elastic's `meta` has lots of restrictions https://www.elastic.co/guide/en/elasticsearch/reference/8.11/mapping-field-meta.html so we can't just dump configuration there. The reason we're using `subType` at all (instead of specifying the subType in the schema overrides) is to leverage syncing to set the subType if possible.

I think your ask here is to have an API for allowing highlight configuration which we're trying to avoid with this PR. Not that we will never have it but we're trying to get away with no configuration whatsoever if we can for now.

positiveVibes4all: I am thinking we could just add fragmentCount and also fragmentSize to meta, I get that we are limited to 5 based on the documentation, but this just seems more flexible in general to rely on the mappings for these fields instead of the schema. I understand though if we do not want to based on the limitations of the meta field.

stellarhoof: Yeah `subType` is ok because it's intrinsic to the field and it's a small payload. If we're going to allow highlighting configuration there should ideally be one way of doing it and `meta` cannot accommodate for arbitrary configuration.

packages/provider-elasticsearch/src/example-types/results/highlighting/merging.js
daedalus28: https://lodash.com/docs/4.17.15#defaultsDeep

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
daedalus28: You can spread iterators directly without Array.from
`_.map(_.join('.'), [...product])`

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
daedalus28: My motivation for adding tests is so each method can be more easily reviewed (and refactorings can be more easily suggested since they can be tested independently)

packages/provider-elasticsearch/src/utils/futil.js
daedalus28: Yup, that's right. We've done this in other places by composing `_.toPairs` with `_.groupBy`

packages/provider-elasticsearch/src/example-types/results/highlighting/README.md
```diff
@@ -0,0 +1,309 @@
...
+2. Transforming highlighted fragments in the elastic response into a structure similar to that of `_source`.
+3. Merging such structure into a hit's `_source`.
+
+## 1. Request
+
+### Fields sent for highlighting
+
+We assume that users want to highlight all the fields present in the query. The most logical approach is to extract relevant fields from the query and send them for highlighting, but for simplicity's sake we send every field in the schema, with some caveats.
+
+#### 1. Sub-fields
```
dubiousdavid: This should be `multi-fields` to be precise. https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

packages/provider-elasticsearch/src/example-types/results/highlighting/README.md
```diff
@@ -0,0 +1,309 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+</details>
+
+#### 2. Fields groups
```
dubiousdavid: This should be `group field` to be precise. https://www.elastic.co/guide/en/elasticsearch/reference/current/copy-to.html

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
```diff
@@ -0,0 +1,204 @@
...
+    it('should merge fragments', () => {
+      const hit = {
+        _source: {
+          name: 'John Wayne',
+          state: 'New Jersey',
+          city: { street: 'Jefferson Ave' },
+        },
+        highlight: {
+          state: ['<em>New</em> Jersey'],
+          'state.exact': ['New <em>Jersey</em>'],
```
dubiousdavid: It might be helpful to have less contrived test data.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,66 @@
...
+  _.find((k) => field.startsWith(k), getArrayFieldsPaths(schema))
+
+const lastWordRegex = /\.(\w+)$/
+
+const getMultiFieldName = (schema, field) => {
+  const [multi, sub] = field.split(lastWordRegex)
+  return schema.fields[multi]?.elasticsearch?.fields?.[sub] ? multi : field
+}
+
+export const transformHighlightResponse = (schema, config, hit) => {
```
dubiousdavid: It might be nice to add TS doc comments for more complicated functions. Also, since we're not using Typescript here documenting the shape of the inputs and outputs in the comment would also be useful.

packages/provider-elasticsearch/src/utils/futil.js
stellarhoof: There would still be this util though but with a different implementation because it's not just `_.toPairs` and `_.groupBy`, after grouping there's `_.mapValues` to get rid of the pairs. And then you have an implementation that is probably as verbose as this one but less efficient.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: Added unit tests for these fns

packages/provider-elasticsearch/src/utils/futil.js
stellarhoof: Do you have a better idea?

packages/provider-elasticsearch/src/example-types/results/type.d.ts
daedalus28: Based on our conversation, we should default this on. The performance concern would be for searches not powering a UI, but there's precedent for thinking about this with flags like `forceFilterOnly`.

Some API cases:

`highlight: {disable: true }` disables highlighting
`highlight: undefined` (or missing) would highlight
`highlight: { /*... config ...*/ }` would highlight with config

packages/provider-elasticsearch/src/example-types/results/highlighting/schema.test.js
```diff
@@ -0,0 +1,82 @@
...
+          subfield: {
+            dataType: 'text',
+          },
+        },
+      },
+    },
+  },
+}
+
+// Bogus test
```
daedalus28: add some explanation on why this is here

packages/provider-elasticsearch/src/schema.js
```diff
@@ -33,13 +33,17 @@ let fromEsIndexMapping = (mapping) => {
           'fields',
           _.flow(
             flatten,
-            F.mapValuesIndexed(({ type, fields }, field) => ({
+            F.mapValuesIndexed((mapping, field) => ({
               field,
               label: _.startCase(field),
               elasticsearch: F.compactObject({
-                dataType: type,
+                ...mapping,
```
daedalus28: _if_ this is truly necessary, we should not spread `mapping` and should wrap it in a field.

Ideally, we process the mapping here into something simpler that the rest of the code can read from

packages/provider-elasticsearch/src/utils/futil.js
```diff
@@ -1,6 +1,12 @@
 import _ from 'lodash/fp.js'
 import F from 'futil'
 
+export const areArraysEqual = _.flow(_.xor, _.isEmpty)
```
daedalus28, line 4: rename to `isArrayEqual`

packages/provider-elasticsearch/src/utils/futil.js
```diff
@@ -1,6 +1,12 @@
...
+
+export const removePrefix = _.curry((prefix, str) =>
```
daedalus28: After our conversation, this doesn't seem helpful. It's not really removing the prefix and is actually just removing characters from the beginning of a string - which is what `slice` is for.

packages/provider-elasticsearch/src/utils/futil.js
daedalus28: This seems fine since alternatives would be less performance and this is a library method

packages/provider-elasticsearch/src/example-types/results/index.test.js
```diff
@@ -1,461 +0,0 @@
...
-    await resultsTest(node, [
-      _.extend(expectedCalls[0], {
-        sort: {
-          [sortField]: 'desc',
-        },
-      }),
-    ])
-  })
-
-  it('Should hoist from tree based on demarcation for hoisting from aggs', () => {
```
daedalus28, line 269: add tests for unsetOnTree to safely remove these

packages/provider-elasticsearch/src/example-types/results/index.test.js
```diff
@@ -1,461 +0,0 @@
...
-          },
-        },
-      ],
-    }
-    let result = { result: input, removed: hoistOnTree(input) }
-    expect(result).toEqual(output)
-  })
-
-  it('Should hoist from tree based on demarcation for hoisting from filters', () => {
-    let input = {
```
daedalus28, line 362: same as above

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
```diff
@@ -0,0 +1,104 @@
+import _ from 'lodash/fp.js'
+import F from 'futil'
+
+export const findByPrefix = (str, arr) =>
+  _.find((k) => _.startsWith(k, str), arr)
+
+export const isLeafField = (field) => !!field?.elasticsearch?.dataType
+
+export const isBlobField = (field) =>
+  field?.elasticsearch?.meta?.subType === 'blob' && isLeafField(field)
```
daedalus28: Based on our conversation, it doesn't seem like we should put this in `meta`. We will always have data that needs to be added to schemas that can't be handled automatically (e.g. by syncing code), so this just adds another place for things to live and makes it more complicated.

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
```diff
@@ -0,0 +1,104 @@
...
+    (acc, field, arrayPath) => {
+      if (isArrayOfObjectsField(field)) {
+        acc[arrayPath] = _.filter(_.startsWith(`${arrayPath}.`), fieldsPaths)
+      }
+      return acc
+    },
+    {},
+    schema.fields
+  )
+}, _.get('elasticsearch.index'))
```
daedalus28, line 40: This can be a flow:

```js
let isChildOf = (path) => _.startsWith(`${path}.`)

_.flow(
  _.pickBy(isArrayOfObjectsField),
  F.mapValuesIndexed((f, path) => _.filter(isChildOf(path), fieldsPaths)),
)


```

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+
+/*
+ * Add given paths to source with includes/excludes lists. Paths get added to
+ * source.includes and removed from source.excludes as necessary.
+ *
+ * Returns added paths.
+ */
+export const addPathsToRequestSource = (schema, source, paths) => {
+  // There's nothing to add.
+  if (_.isEmpty(paths) || _.isEmpty(F.omitBlank(source))) {
```
daedalus28: `F.isBlankDeep`

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+
+  // With wildcards expanded.
+  const expanded = {
+    includes: expandGlobs(schema, source.includes),
+    excludes: expandGlobs(schema, source.excludes),
+  }
+
+  // To understand this, visualize a Venn diagram with three intersecting sets
+  // one for each of includes, excludes, and paths.
+  const pathsToAdd = _.union(
```
daedalus28: this is confusing because the prop `paths`  was called `pathsToAdd` in the previous method, which means this is paths that will be added based on the paths that need to be added 😕

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+    _.isEmpty(source.includes) ? [] : _.difference(paths, expanded.includes)
+  )
+
+  // There's nothing to add.
+  if (_.isEmpty(pathsToAdd)) {
+    return []
+  }
+
+  const withAddedPaths = F.omitBlank({
+    includes: _.union(pathsToAdd, expanded.includes),
```
daedalus28: please flip the args here for consistency with the below :)

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+  const shouldAddPaths = (key) =>
+    !areArraysEqual(expanded[key], withAddedPaths[key])
+
+  if (!_.isEmpty(source.includes) && shouldAddPaths('includes')) {
+    source.includes = withAddedPaths.includes
+  }
+
+  if (shouldAddPaths('excludes')) {
+    source.excludes = withAddedPaths.excludes
+  }
```
daedalus28: experiment with changing the logic to handle include/exclude without a union and see if that's clearer

daedalus28: and if not, add a comment explaining why you have a check with `shouldAddPaths`

packages/provider-elasticsearch/src/example-types/results/highlighting/search.js
```diff
@@ -0,0 +1,44 @@
...
+  pre: '<b class="search-highlight">',
+  post: '</b>',
+}
+
+export const searchWithHighlights = (node, search, schema) => async (body) => {
+  // Paths for fields to always include regardless of whether the user included
+  // them. They will be removed from the response hits so there's no harm done.
+  const pathsToAdd = _.flatten(_.values(getArrayOfObjectsPathsMap(schema)))
+
+  // body._source is mutated here
```
daedalus28: let's avoid mutation here

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+  return _.flatMap((field) => {
+    const copy_to = field.elasticsearch?.copy_to
+    if (_.isEmpty(copy_to)) return copy_to ?? []
+    const product = new CartesianProduct(copy_to, subFieldsNames)
+    return [...(copy_to ?? []), ..._.map(_.join('.'), [...product])]
+  }, schema.fields)
+}, _.get('elasticsearch.index'))
+
+const isFieldsGroupPath = (schema, path) =>
+  !!findByPrefix(path, getHighlightFieldsGroupsPaths(schema))
```
daedalus28: we do not not not need not not 🙃

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+  _.keys(_.pickBy('highlight', schema.elasticsearch?.subFields))
+
+/*
+ * Paths of all fields groups and their subfields that can be highlighted.
+ */
+export const getHighlightFieldsGroupsPaths = _.memoize((schema) => {
+  const subFieldsNames = getHighlightSubFieldsNames(schema)
+  return _.flatMap((field) => {
+    const copy_to = field.elasticsearch?.copy_to
+    if (_.isEmpty(copy_to)) return copy_to ?? []
```
daedalus28: this can be reduced to
```js
if (_.isEmpty(copy_to)) return []
```

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+    const product = new CartesianProduct(copy_to, subFieldsNames)
+    return [...(copy_to ?? []), ..._.map(_.join('.'), [...product])]
```
daedalus28: `return [...copy_to, ..._.map(_.join('.'), product)]`

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+
+/*
+ * Paths of all fields groups and their subfields that can be highlighted.
+ */
+export const getHighlightFieldsGroupsPaths = _.memoize((schema) => {
+  const subFieldsNames = getHighlightSubFieldsNames(schema)
+  return _.flatMap((field) => {
+    const copy_to = field.elasticsearch?.copy_to
+    if (_.isEmpty(copy_to)) return copy_to ?? []
+    const product = new CartesianProduct(copy_to, subFieldsNames)
```
daedalus28: do the spread here so that knowledge of combinatorics js is only on this line

daedalus28: rename product -> subFieldPairs (or something better to communicate what it is)

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
daedalus28: also, I think you can just use `_.find` instead of `findByPrefix` since the original path should be in there since you're including `copy_to` in `getHighlightFieldsGroupsPaths`

daedalus28: Furthermore, you might want to consider a simpler list of field groups by just pulling out and doing a unique union on `copy_to` without worrying about sub fields for this method

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+  const fieldGroupsInQuery = F.reduceTree()(
+    (acc, val, key) =>
+      isFieldsGroupPath(schema, val)
+        ? F.push(val, acc)
+        : isFieldsGroupPath(schema, key)
+        ? F.push(key, acc)
+        : acc,
+    [],
+    node._meta?.relevantFilters
+  )
```
daedalus28: This can be made clearer by factoring out a `collectKeysAndValues` method (ideally come up with a noun that represents keys and values - e.g. scalars although this method could work on non-scalars)

```js
let collectKeysAndValues = (f, tree) =>
  F.reduceTree()(
    (acc, val, key) =>
      f(val) ? F.push(val, acc) : f(key) ? F.push(key, acc) : acc,
    [],
    tree,
  )

const fieldGroupsInQuery = collectKeysAndValues(
  (x) => isFieldsGroupPath(schema, x), // curry to simplify
  node._meta?.relevantFilters,
)
```

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const queryStr = JSON.stringify(node._meta?.relevantFilters)
```
daedalus28: Let's just add a comment explaining that stringification is a more performant alternative to tree walking to replace fields/values.

Also just a warning - this isn't always safe since you lose type information (e.g. `Date` objects will come back as strings). In this case, I think it's safe since elastic queries are stringified over the wire at some point - but double check to be sure.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -0,0 +1,180 @@
...
+      const regexp = new RegExp(_.join('|', pathsToReplace), 'g')
+      return JSON.parse(_.replace(regexp, path, queryStr))
+    }
+  }
+
+  return F.mapValuesIndexed(
+    (field, path) =>
+      F.omitBlank({
+        fragment_size: isBlobField(field) ? 250 : null,
+        number_of_fragments: isBlobField(field) ? 3 : null,
```
daedalus28: let's factor out `blobFragment: { fragment_size: 250, number_of_fragments: 3 }` and then spread it on if it's a blob field so that the arbitrary config is clear at the top of the file.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,180 @@
...
+ * `city.subfield` will be grouped under `city`.
+ */
+const groupByMultiField = _.curry((schema, highlight) =>
+  groupByIndexed((v, path) => {
+    const [multi, sub] = path.split(lastWordRegex)
+    return schema.fields[multi]?.elasticsearch?.fields?.[sub] ? multi : path
+  }, highlight)
+)
+
+export const transformResponseHighlight = (
```
daedalus28: add a comment to explain this

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
```diff
@@ -0,0 +1,607 @@
...
+          '<em>Imperial</em> College <em>London Abdus Salam</em> Library',
+      })
+    })
+  })
+
+  describe('blob text fields', () => {
+    it('should not merge fragments', () => {
+      const hit = {
+        highlight: {
+          'library.about': [
```
daedalus28, line 143: Consider this case:

```js
input = {
  'library.about': [
    'The <em>Abdus</em> Salam Library is',
    'is the <em>largest</em> of',
  ],
  'library.about.subfield': [
    'The Abdus <em>Salam Library</em> is',
    'went on to <em>form part</em> of',
    'used <em>extensively</em> by members of the college',
  ],
}

}
output = {
  'library.about': [
    'The <em>Abdus Salam Library</em> is',
    'is the <em>largest</em> of',
    'went on to <em>form part</em> of',
    'used <em>extensively</em> by members of the college',
  ],
}

```

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
stellarhoof: We decided to set these flags in the field schema instead of elastic's `meta`

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
```diff
@@ -0,0 +1,607 @@
...
+          },
+        },
+        highlight: {
+          'library.books.cover.title': [
+            'Nineteen <em>Eighty-Four</em>',
+            '<em>The</em> Great Gatsby',
+          ],
+        },
+      }
+      let nestedArrayIncludes = { 'library.books': ['cover.author'] }
```
dubiousdavid: This might be better as an array of paths that are array agnostic: 
```
['library.books.cover.author']
```

packages/provider-elasticsearch/src/example-types/results/highlighting/search.js
```diff
@@ -0,0 +1,48 @@
...
+    highlight: {
+      pre_tags: [tags.pre],
+      post_tags: [tags.post],
+      number_of_fragments: 0,
+      fields: getRequestHighlightFields(schema, node),
+    },
+  })
+
+  for (let hit of response.hits.hits) {
+    let nestedArrayIncludes = node.highlight?.nestedArrayIncludes
```
dubiousdavid: Might make sense to change `nestedArrayIncludes` to `copySourceFields` given the path feedback.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
dubiousdavid: Alejandro and I discussed this and weren't sure if this was absolutely necessary. @daedalus28 can you please clarify if this is just a nice-to-have or absolutely needed?

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
stellarhoof: Outdated

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: Outdated

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
daedalus28: It's not necessary for now, but unless solving this use could potentially unify some code paths (making it simpler and more powerful at the same time)

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
daedalus28: this was marked as resolved, but I don't see it changed

packages/react/src/utils/futil.js
```diff
@@ -24,3 +24,21 @@ export let aspectWrapper = F.aspect({
...
+    : F.reduceIndexed(
+        (output, value, key) =>
+          _.merge(output, fn(value, F.dotJoinWith(F.isNotNil)([paths, key]))),
+        {},
+        input
+      )
+)
+
+// { a: [{ b: { c: 1 } }] } => { a: [{ 'b.c': 1 }] }
+export const flattenObjectsNotArrays = flattenObjectWith((value, path) => {
```
daedalus28, line 40: Given that this is in other parts of our code base (outside contexture, I think), it's probably time to think about promoting these to futil properly (in a followup PR)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+}
+
+/*
+ * Merge elastic hit highlights onto hit source.
+ *
+ * As a clever developer, you will notice that the following function is a dirty
+ * and unholy version `_.merge`. So before you refactor it to use exactly that,
+ * consider that this implementation is about 100x faster than (immutable)
+ * `_.merge`. Query 100 records with arrays of thousands of elements each and
```
daedalus28: `_.merge` does not have to be immutable. In lodash/fp, it is by default - but we can include the mutable version here instead. Given the size and complexity of this method, it's probably worth doing so.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/**
+ * Remove each path in `paths` from `hit._source`.
+ */
+export let removePathsFromSource = (schema, hit, paths) => {
+  // Nothing to do
+  if (_.isEmpty(paths)) return
+
+  // "aoo" stands for "array of objects", because I was tired of typing it out
```
daedalus28: I get it, but this code is harder to follow as a result. Let's also be careful of our comment tone (especially in OSS repos)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
daedalus28: Let's also be careful of our comment tone (especially in OSS repos)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+        removeEntireArray ||
+        _.map((path) => path.slice(aooPath.length + 1), paths)
+      )
+    }, _.groupBy(getAooPath, aooPaths)),
+  }
+
+  let removePathsFromArray = (paths) => (arr) =>
+    _.reduce(
+      (acc, item) => {
+        for (let path of paths) F.unsetOn(path, item)
```
daedalus28: [`F.popProperty`](https://smartprocure.github.io/futil-js/#popProperty) will return the removed item and might clean this up. We're also doing something very similar with hoisting in this repo using `unsetOnTree` (https://github.com/smartprocure/contexture/blob/main/packages/provider-elasticsearch/src/utils/futil.js#L4)

daedalus28: In fact, I think much of this part of the code might just be unsetOnTree

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+ */
+export let transformResponseHighlight = (
+  schema,
+  hit,
+  tags,
+  nestedArrayIncludes = {}
+) => {
+  let arrayOfObjectsPaths = _.keys(getArrayOfObjectsPathsMap(schema))
+
+  let getIndexedArrayObject = (arr, fragments, arrayPath, itemPath) => {
```
daedalus28: This name doesn't communicate it's purpose and there's no comments to elaborate either

daedalus28: Let's also add comments explaining why this is so performance critical (since it runs on every result)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+    let fragmentsMap = _.groupBy(stripTags(tags), fragments)
+    return F.reduceIndexed(
+      (acc, item, index) => {
+        let fragments = fragmentsMap[itemPath ? _.get(itemPath, item) : item]
```
daedalus28: this is `F.getOrReturn`

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+        if (fragments) {
+          F.setOn(
+            F.dotJoin([`${index}`, itemPath]),
```
daedalus28: add a comment about why we're casting to a string (to avoid compacting 0 indexes, but still wanting to omit an empty itemPath)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+    _.mapValues(_.flatten),
+    F.reduceIndexed((acc, fragments, path) => {
+      let arrayPath = getArrayPath(path)
+      if (arrayPath) {
+        acc[arrayPath] = getIndexedArrayObject(
+          acc[arrayPath] ?? {},
+          fragments,
+          arrayPath,
+          path.slice(arrayPath.length + 1)
+        )
```
daedalus28: If we group by array path first, we can separate this into small individual steps to make it easier to follow (and maybe even avoid a reduce)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
daedalus28: We spoke and this comment is actually incorrect - so let's update the comment accordingly

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+
+  // "aoo" stands for "array of objects", because I was tired of typing it out
+  // over and over again.
+  let aooMap = getArrayOfObjectsPathsMap(schema)
+  let allAooPaths = _.keys(aooMap)
+  let getAooPath = (path) => findByPrefix(path, allAooPaths)
+  let [aooPaths, otherPaths] = _.partition(getAooPath, paths)
+
+  let toRemove = {
+    ...F.arrayToObject(_.identity, _.constant(true), otherPaths),
```
daedalus28: this is [`F.flags`](https://smartprocure.github.io/futil-js/#flags)

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
```diff
@@ -0,0 +1,189 @@
...
+    }, {})
+  )(hit.highlight)
+
+  if (!_.isEmpty(highlight)) hit.highlight = highlight
+}
+
+/**
+ * Remove each path in `paths` from `hit._source`.
+ */
+export let removePathsFromSource = (schema, hit, paths) => {
```
daedalus28, line 141: Based on our discussion, you can likely substantially simplify this by thinking of it in terms of an `omit`. To handle the case that made this complicated (omitting all paths from an array of objects), you can determine that up front before iterating to omit.

The idea here is that you're reframing the problem up front to make the subsequent code much simpler.

daedalus28 (changes requested):

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
dubiousdavid: Even with a comment this is a good example of code that is hard to follow because it requires you to know that `F.dotJoin` also does compaction. Something that is not obvious from the name `dotJoin`.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.test.js
stellarhoof: Merging these types of fragments does not unify any code paths, on the contrary, we'd need to write code to handle this case.

packages/provider-elasticsearch/src/example-types/results/highlighting/README.md
stellarhoof: Sub-field is correct here, we send multi-fields but we also send sub-fields (ex: `book.title.subfield`)

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: > [F.popProperty](https://smartprocure.github.io/futil-js/#popProperty) will return the removed item and might clean this up.

I don't need to return the popped value

> In fact, I think much of this part of the code might just be unsetOnTree

It could, but this fn is more complicated for performance reasons and `unsetOnTree` looks pretty expensive. I'll do post-merge performance test using `unsetOnTree`.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: `F.flags` unflattens the paths: `F.flags(['a.b']) => { a: { b: true } }` whereas I want `{ a.b: true }`

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
stellarhoof: Done

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: Reworded the comment to be less flairy and more to the point and fixed it to clarify that I had used a mutating merge.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: As part of cleaning up the code some, this doesn't exist anymore.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: This fn is simpler now.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: This code doesn't exist anymore.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: Done

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: This code doesn't exist anymore.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: Done.

packages/provider-elasticsearch/src/example-types/results/highlighting/response.js
stellarhoof: Reworked the code and now it's simpler

dubiousdavid (approved):

PR #161, opened 2023-11-14
Fix regression in PR #159

stellarhoof:
Fix regression introduced in https://github.com/smartprocure/contexture/pull/159 where the `in` operator fails if the right-hand side is `undefined`

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/highlighting.js
+++ b/packages/provider-elasticsearch/src/example-types/results/highlighting.js
@@ -144,7 +144,7 @@ let handleNested = ({
 
   if (filterNested) {
     for (const path of nested) {
-      if (!(path in hit.highlight)) {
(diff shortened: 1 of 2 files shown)


packages/provider-elasticsearch/src/example-types/results/highlighting.js
```diff
@@ -144,7 +144,7 @@ let handleNested = ({
 
   if (filterNested) {
     for (const path of nested) {
-      if (!(path in hit.highlight)) {
+      if (!_.has(path, hit.highlight)) {
```
positiveVibes4all, line 147: I tested both of these just to be sure and they are equivalent, what does this fix?

stellarhoof: `"foo" in undefined` fails

PR #163, opened 2023-11-16
Allow filtering facet options when they are ObjectIDs

stellarhoof:
- Split facet filter query on whitespace to fix regression introduced in https://github.com/smartprocure/contexture-mongo/pull/51 where a facet filter string would get split at alphabetical boundaries (ex: `"ab56ff"` -> `["ab", "56", "ff"]`).
- Cast field we're filtering on in the facet to a string before matching so we can apply regex to ObjectIDs as well

image

### DIFF
--- a/packages/provider-mongo/src/example-types/facet.js
+++ b/packages/provider-mongo/src/example-types/facet.js
@@ -23,12 +23,12 @@ let sortAndLimitIfNotSearching = (should, limit) =>
   sortAndLimitIfSearching(!should, limit)
 
 let getSearchableKeysList = _.flow(
-  _.getOr('_id', 'label.fields'),
+  _.getOr('_strfield', 'label.fields'),
   _.castArray,
-  _.map((label) => (label === '_id' ? label : `label.${label}`))
+  _.map((label) => (label === '_strfield' ? label : `label.${label}`))
 )
 
-let getMatchesForMultipleKeywords = (list, optionsFilter) => ({
+let getMatchesForMultipleKeywords = (list, filterWords) => ({
   $and: _.map(
     (option) => ({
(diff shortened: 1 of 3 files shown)


packages/provider-mongo/src/example-types/facet.js
```diff
@@ -41,30 +41,33 @@ let getMatchesForMultipleKeywords = (list, optionsFilter) => ({
...
+}
 
-let mapKeywordFilters = (node) =>
-  node.optionsFilter &&
-  _.flow(getSearchableKeysList, (list) => ({
-    $match: setMatchOperators(list, node),
-  }))(node)
+let mapKeywordFilters = (node) => [
+  // Cast field we're searching on to string to enable regex matches on it.
+  { $addFields: { _strfield: { $toString: '$_id' } } },
```
daedalus28: We might want a better name than `_strfield` that communicates that it's a string version of `_id`, e.g. `_idString`

daedalus28 (changes requested):
Great work 👏

packages/provider-mongo/src/example-types/facet.js
stellarhoof: The reason I didn't reference `id` in it is because it's not necessarily one. It's called `_id` because that's what we grouped on: https://github.com/smartprocure/contexture/blob/0c25aca67dbe4ffedf5d1aab23b5d92805bebbd2/packages/provider-mongo/src/example-types/facet.js#L152

We could name it `_nodeFieldStr` or something

daedalus28: Works for me

daedalus28 (approved):

PR #164, opened 2023-11-17
[POC] Move example types into their own package

stellarhoof:
Proof of concept PR to explore whether it's feasible to move example types into their own package.

Copied types `bool` and `exists` from both the client and the mongo provider to a new package `packages/example-types`. Also tracked the built code under `build` so we can see what's being generated.

I think this could work. We can get nicely encapsulated example types under `src` and also per-package output in `build`. The coolest thing is that shared code gets output into their own chunks so there's no risk of importing the same shared code multiple times.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ node_modules/
 
 # Build output
 dist
-!packages/example-types/dist
 
 # Jest
 coverage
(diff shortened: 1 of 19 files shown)


stellarhoof:
We may come back to this later.

PR #168, opened 2023-11-20
Remove minio for testing

stellarhoof:
- Use `mongo-memory-server` instead of `minio` in tests
- Remove peer dependency on mongodb. Mongodb was only being used for `ObjectId`, which we can get from `bson` instead to remove this dependency.


jest.config.js
```diff
@@ -29,12 +29,5 @@ export default {
       displayName: 'provider-mongo',
       testMatch: ['<rootDir>/packages/provider-mongo/src/**/*.test.js'],
     },
-    // TODO: Fix tests
-    // {
-    //   displayName: 'provider-mongo:integration',
-    //   testMatch: [
-    //     '<rootDir>/packages/provider-mongo/integration-test/example-types/**/*.js',
-    //   ],
-    // },
```
stellarhoof, line 38: Made a regular test in this PR

packages/provider-mongo/src/index.test.js
stellarhoof, line 1: This was the integration test

packages/client/src/index.test.js
```diff
@@ -897,7 +897,7 @@ let AllTests = (ContextureClient) => {
     tree.mutate(['root', 'a'], { values: [1] })
     let node = tree.getNode(['root', 'b'])
     // Allow updating to start (after debounce elaspses) but before the service finishes
-    await Promise.delay(5)
+    await Promise.delay(10)
```
daveyarwood, line 900: Sleeping (or its moral equivalent here) in tests is usually a sign of flaky tests, and should generally be avoided.

I wrote [a blog post](https://blog.djy.io/fixing-timing-issues-in-tests/) a while back about a pragmatic approach to fixing timing issues in tests. It isn't ideal, and sometimes, there are better ways to approach the tests you want to write, but at the very least, it's better than attempting to find the right magic number of seconds to sleep to get the tests passing and driving the total amount of time for your test runs up.

(We were already sleeping before this PR, so I don't think this is critical to fix right now, but it's something we should keep in mind going forward.)

stellarhoof: I don't love this either and as a matter of fact, I had to bump this delay precisely because these tests are flaky. I may come back at some point to try and improve them but they're not high priority right now.

Kudos on writing a blog post and everything!

rudolph9 (approved):

PR #171, opened 2023-11-29
Use non-analyzed field for terms_stats value_field

stellarhoof:
TBD...


stellarhoof:
Abandoning in favor of using the non-legacy example types.

PR #182, opened 2024-01-12
[develop] Fix Facet toasts with [object object]

quinn895:
This PR fixes an issue with the toasts showing [object object]. The reason is because the display() callback function can sometimes return a React node instead of a string. This PR fixes it so that it will get the safe string.

### DIFF
--- a/.changeset/react-facet-title-in-label.md
+++ b/.changeset/react-facet-title-in-label.md
@@ -1,0 +1,5 @@
+---
+'contexture-react': patch
+---
+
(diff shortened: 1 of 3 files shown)


stellarhoof (approved):
Looks good 👍

stellarhoof:
@quinn895 A couple of comments:

- No need to add `[develop]` to the PR title. Actually there's no `develop` branch in this repo.
- Try not to refer to `spark` since that's a private repo and this is an open source repo.

PR #189, opened 2024-01-29
Elasticsearch provider: only highlight text fields

stellarhoof:
Remove `subFields` configuration from the schema. Instead only send `text` fields for highlighting. This both simplifies the API and reduces payload to elastic, as well as fixing an issue where top-level `keyword` fields were being highlighted when they should not be.

Also improve performance (from ~70ms to ~10ms on a page size of 100 with large arrays each) when handling elastic responses.

### DIFF
--- a/.changeset/wet-maps-shout.md
+++ b/.changeset/wet-maps-shout.md
@@ -1,0 +1,8 @@
+---
+'contexture-elasticsearch': minor
+---
+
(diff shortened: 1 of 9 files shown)


stellarhoof:
I'm merging this PR as-is because it's a fix needed in the main product. I'll address any feedback in a follow-up.

PR #196, opened 2024-02-05
Low-hanging perf improvements in highlighting code

stellarhoof:
Use `F.pushOn` in more places and a few perf tweaks in hot code paths.

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/metricGroups/percentilesGroupStats.js
+++ b/packages/provider-elasticsearch/src/example-types/metricGroups/percentilesGroupStats.js
@@ -6,7 +6,7 @@ import { groupStats } from './groupStatUtils.js'
 // [1, 2, 3] -> [{to: 1}, {from: 1, to: 2}, {from: 2, to: 3}, {from: 3}]
 let boundariesToRanges = _.flow(
   F.mapIndexed((to, i, list) => F.compactObject({ from: list[i - 1], to })),
-  (arr) => F.push({ from: _.last(arr).to }, arr)
+  (arr) => F.pushOn(arr, { from: _.last(arr).to })
 )
 
 let drilldownToRange = (drilldown) => {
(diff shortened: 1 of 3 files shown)


dubiousdavid (commented):
It may be worth looking at some of these functions/flows to determine if a simple `for` loop could be much more performant.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -95,16 +95,19 @@ let getFieldSubFields = (field) =>
...
-    {},
-    schema.fields
-  )
+let getSchemaSubFields = _.memoize(
+  (schema) =>
+    F.reduceIndexed(
+      (acc, field, path) => {
+        let subFields = getFieldSubFields(field)
+        for (let k in subFields) acc[`${path}.${k}`] = subFields[k]
+        return acc
```
dubiousdavid: Mutating the accumulator and then returning it feels kind of silly to me. Why not just use a `for` loop. It's easy to read and clearly indicates the imperative approach you're taking here.

packages/provider-elasticsearch/src/example-types/metricGroups/percentilesGroupStats.js
```diff
@@ -6,7 +6,7 @@ import { groupStats } from './groupStatUtils.js'
 // [1, 2, 3] -> [{to: 1}, {from: 1, to: 2}, {from: 2, to: 3}, {from: 3}]
 let boundariesToRanges = _.flow(
   F.mapIndexed((to, i, list) => F.compactObject({ from: list[i - 1], to })),
-  (arr) => F.push({ from: _.last(arr).to }, arr)
+  (arr) => F.pushOn(arr, { from: _.last(arr).to })
```
dubiousdavid: I think I actually prefer `F.push` here. Mixing non-mutating with mutating in the same flow is kind of awkward for me. If you want to mutate an array with `push` you should just loop over the array and do that.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -148,7 +151,7 @@ export let getAllHighlightFields = _.memoize((schema) => {
 let collectKeysAndValues = (f, coll) =>
   F.reduceTree()(
     (acc, val, key) =>
-      f(val) ? F.push(val, acc) : f(key) ? F.push(key, acc) : acc,
+      f(val) ? F.pushOn(acc, val) : f(key) ? F.pushOn(acc, key) : acc,
```
dubiousdavid: Same kind of thing here. Using `reduce` with a mutating call in the reducing function feels a little awkward to me. Why not just walk the tree and and `push` onto an array?

dubiousdavid (approved):

PR #197, opened 2024-02-06
Add test script to every package

stellarhoof:
I always forget how to run the tests and have to rely on my shell's history.


rudolph9 (approved):

dubiousdavid (commented):
Can you please explain what each piece of the test script actually does? Also, why are they not all the same?

stellarhoof:
@dubiousdavid
- `NODE_OPTIONS=--experimental-vm-modules` is so jest can load es modules natively without resorting to babel to transpile source.
- `NODE_NO_WARNINGS=1` is to suppress the node warning resulting from using the above option.
- The scripts in the packages use `yarn run -T` to run `jest` installed at the top-level since they don't install jest themselves.

PR #200, opened 2024-02-14
Fix error when there are no results node includes

stellarhoof:
A results' node `include` property may be empty, which means include everything.

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
+++ b/packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
@@ -190,7 +190,7 @@ export let getRequestHighlightFields = (schema, node) => {
   // TODO: `highlightOtherMatches` is an undocumented configuration value that we
   // are currently using to work around performance issues when highlighting
   // fields not included in the node.
-  if (!node.highlight?.highlightOtherMatches) {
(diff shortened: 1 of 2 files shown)


packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -190,7 +190,7 @@ export let getRequestHighlightFields = (schema, node) => {
   // TODO: `highlightOtherMatches` is an undocumented configuration value that we
   // are currently using to work around performance issues when highlighting
   // fields not included in the node.
-  if (!node.highlight?.highlightOtherMatches) {
+  if (!node.highlight?.highlightOtherMatches && _.isArray(node.include)) {
```
dubiousdavid, line 193: Small thing, but we should probably prefer the `Array.isArray` built-in here.

stellarhoof: I agree in general but I think we should keep the pattern consistent and switch to it everywhere once.

dubiousdavid: There shouldn't be any harm here, but okay.

dubiousdavid (approved):

PR #202, opened 2024-03-08
Update datemath expressions

daedalus28:
Fixes date math expressions that were missing rounding and 1ms offset. Should improve caching and fix subtle bugs.


packages/provider-elasticsearch/src/utils/dateUtil.js
```diff
@@ -27,21 +27,22 @@ let quarterToOffset = {
...
-  last90Days: { from: 'now-90d', to: 'now' },
-  last180Days: { from: 'now-180d', to: 'now' },
-  last12Months: { from: 'now/d-12M', to: 'now' },
-  last15Months: { from: 'now/d-15M', to: 'now' },
-  last18Months: { from: 'now/d-18M', to: 'now' },
-  last24Months: { from: 'now/d-24M', to: 'now' },
-  last36Months: { from: 'now/d-36M', to: 'now' },
-  last48Months: { from: 'now/d-48M', to: 'now' },
-  last60Months: { from: 'now/d-60M', to: 'now' },
+  last1Hour: { from: 'now/h-1h', to: 'now/h-1ms' },
```
dubiousdavid, line 32: I'm not sure this is what we want. The last hour no longer means one hour from right now. In general, these changes are not inclusive of the current time which would be misleading for something like bids where you wouldn't want to have to check at the start of the next hour or the next day for something that was actually present right now.

PR #203, opened 2024-03-15
Allow dots in keywords when searching on exact words

stellarhoof:
Allow dots in keywords tags when matching on exact words

https://github.com/smartprocure/contexture/assets/4336260/c44b7dc6-453a-4d60-9b6d-781f068906ec

### DIFF
--- a/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -18,6 +18,10 @@ import ActionsMenu from '../TagsQuery/ActionsMenu.js'
 import { useOutsideClick } from '@chakra-ui/react-use-outside-click'
 import { sanitizeTagInputs } from 'contexture-elasticsearch/utils/keywordGenerations.js'
 import KeywordGenerations from './KeywordGenerations.js'
+import {
+  alphaNumericRegEx,
+  alphaNumericRegExWithDots,
+} from '../../greyVest/utils.js'
 
 let innerHeightLimit = 40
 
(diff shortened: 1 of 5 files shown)


packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -132,7 +136,11 @@ let TagsWrapper = observer(
     popoverOffsetY,
     theme: { Icon, TagsInput, Tag, Popover },
     joinOptions,
-    wordsMatchPattern,
+    // Allow tag keywords to contain dots in them if user is searching for exact
```
conraddavisjr:
```suggestion
    // Allow tag keywords to contain dots in them if the user is searching for exact
```

conraddavisjr (approved):
LGTM! Pending one more ✅ from another reviewer.
I dropped feedback for one minor typo adjustment

packages/react/src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -132,7 +136,11 @@ let TagsWrapper = observer(
     popoverOffsetY,
     theme: { Icon, TagsInput, Tag, Popover },
     joinOptions,
-    wordsMatchPattern,
+    // Allow tag keywords to contain dots in them if the user is searching for exact
+    // words instead of their variations.
+    wordsMatchPattern = node.exact
+      ? alphaNumericRegExWithDots
+      : alphaNumericRegEx,
```
dubiousdavid: Why would we not always use `alphaNumericRegExWithDots`? Also, why are the regexs so strange looking?

stellarhoof: I thought that's we had agreed on, that if you didn't search for exact you'd get the current behavior?

I think they are strange looking because they were copied from lodash. See [this comment](https://github.com/smartprocure/contexture/pull/203/files#diff-431701c69ca84ff7bf09ad61f1da89a10d000c32116db6b3795354611d40de6eR40)

PR #205, opened 2024-04-01
Allow some special characters on TagsQuery component

stellarhoof:
The tags component used in the search bar currently strips out non-word characters in each tag. For example a tag `my+email` will be converted to `my email`. We do this to visually convey tags that contain text that closely resembles what we send to elastic (see https://github.com/smartprocure/contexture-elasticsearch/pull/170 for more information).

There are three places to think about when reviewing this PR:

1. Contexture react: we strip some special characters for presentation purposes
2. Contexture elasticsearch: we strip reserved `query_string` characters before sending the elastic request to prevent our queries from exploding. See https://github.com/smartprocure/contexture/blob/feature/allow-special-keyword-characters/packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js#L36
3. ElasticSearch: text gets stripped out of special characters when analyzed so there is no point in sending such characters in the query. You can see which tokens our analyzers generate by calling `/{index}/_analyze` with a payload such as `{ "analyzer" : "exact", "text" : "student discourse"}`

**Behavior in this PR**

- Strip the following `query_string` reserved characters in the tags component: `| > < ! ( ) { } [ ] ^ " ~ * ? \`.
- Additionally, strip `; , $ '` as well since they are removed by all our analyzers anyway.
- Escape the characters `& + - = : /` in contexture-elasticsearch because those are kept by our analyzers but are also reserved `query_string` characters.

**Further notes**

The props `wordsMatchPattern`, `maxWordsPerTag`, and `maxCharsPerTagWord` taken by the `ExpandableTagsInput` and `TagsInput` component are not passed from anywhere (including spark) so this PR removes them. This is a case of adding props in case they're useful but without a use case.

A common package `contexture-util` was introduced to share code between contexture packages.


packages/react/src/greyVest/TagsInput.js
```diff
@@ -118,19 +105,4 @@ let TagsInput = forwardRef(
   }
 )
 
-// Just uses an internal observable array
-export let MockTagsInput = inject(() => {
```
stellarhoof, line 122: This component was unused. Looks like it was used for a test at some point.

positiveVibes4all:
> The tags component used in the search bar currently strips out non-word characters in each tag. For example a tag `my+email` will be converted to `my email`. We do this to visually convey tags that contain text that closely resembles what we send to elastic (see [smartprocure/contexture-elasticsearch#170](https://github.com/smartprocure/contexture-elasticsearch/pull/170) for more information).
> 
> There are three places to think about when reviewing this PR:
> 
> 1. Contexture react: we strip some special characters for presentation purposes
> 2. Contexture elasticsearch: we strip reserved `query_string` characters before sending the elastic request to prevent our queries from exploding. See https://github.com/smartprocure/contexture/blob/feature/allow-special-keyword-characters/packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js#L36
> 3. ElasticSearch: text gets stripped out of special characters when analyzed so there is no point in sending such characters in the query. You can see which tokens our analyzers generate by calling `/{index}/_analyze` with a payload such as `{ "analyzer" : "exact", "text" : "student discourse"}`
> 
> **Behavior in this PR**
> 
> * Strip the following `query_string` reserved characters in the tags component: `| > < ! ( ) { } [ ] ^ " ~ * ? \`.
> * Additionally, strip `; , $ ' &` as well since they are removed by all our analyzers anyway.
> * Escape the characters `+ - = : /` in contexture-elasticsearch because those are kept by our analyzers but are also reserved `query_string` characters.
> 
> **Further notes**
> 
> The props `wordsMatchPattern`, `maxWordsPerTag`, and `maxCharsPerTagWord` taken by the `ExpandableTagsInput` and `TagsInput` component are not passed from anywhere (including spark) so this PR removes them. This is a case of adding props in case they're useful but without a use case.

I asked Marc for the spreadsheet he made about the analyzers, we might want to update this at some point but at the very least this should live somewhere in our notion documentation: https://docs.google.com/spreadsheets/d/1ObJyl2w0e1jszgYIdk9yUuZU7N9oeSOOtFDFRK_LDwg/edit#gid=0

Here is the code he used to generate this as well: https://github.com/smartprocure/elasticsearch-test-suite/blob/master/src/index.js

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
```diff
@@ -29,11 +29,16 @@ let addQuotesAndDistance = _.curry((tag, text) => {
...
-  // Replace characters with white space ` `
-  _.replace(/([+\-=&|!(){}[\]^"~*?:\\/<>;,$'])/g, ' ')
+  // Most of these characters are `query_string` reserved characters. See
+  // https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters
+  // The characters `;,$'&` are not reserved but they get stripped out by our
+  // analyzers so there's no point in sending them.
+  _.replace(/([&|!(){}[\]^"~*?\\<>;,$'])/g, ' '),
+  // These characters are not stripped out by our analyzers but they are
+  // `query_string` reserved characters so we need to escape them.
+  _.replace(/([+\-=:/])/g, '\\$1')
```
positiveVibes4all: Even though we are stripping out the char '\\', it should not exist in this replace for reserved characters based on the requirements: [found here](https://govspend.atlassian.net/browse/GS-6081)

> Escape + - = & : /

packages/react/src/greyVest/utils.js
```diff
@@ -6,36 +6,47 @@ export let openBinding = (...lens) => ({
...
 )
 
-export let wordRegex = /[-\w]+/g
-export let wordRegexWithDot = /[-.\w]+/g
+export let createTags = ({ input, splitCommas, sanitizeTagFn }) =>
+  _.flow(
+    _.trim,
+    splitCommas ? splitTagOnComma : _.identity,
+    _.castArray,
+    sanitizeTagFn ? _.map(sanitizeTagFn) : _.identity
```
positiveVibes4all: In what case will this be called without a sanitizeTagFn?

packages/react/src/greyVest/TagsInput.js
```diff
@@ -1,18 +1,17 @@
...
+import { createTags } from './utils.js'
 
 let isValidInput = (tag, tags) => !_.isEmpty(tag) && !_.includes(tag, tags)
 
 let TagsInput = forwardRef(
   (
     {
       tags,
-      addTags,
+      addTags: setTags,
```
positiveVibes4all, line 14: I see setTags in the story but nowhere else, where is this used?

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
stellarhoof: In this context it's escaping the `-` character

packages/react/src/greyVest/utils.js
stellarhoof: Full disclaimer: I think `sanitizeTagFn` is a terrible API. Having said that, `sanitizeTagFn` is only currently relevant to transform the tags text for the main search bar, from which we end up sending elastic `query_string` queries.

An example of where it's not needed is [the TagsText component](https://github.com/smartprocure/contexture/blob/main/packages/react/src/exampleTypes/TagsText.js#L37), which I think is used in the "Matches" filter.

packages/react/src/greyVest/TagsInput.js
stellarhoof: They're passed [here](https://github.com/smartprocure/contexture/blob/3658ef6ad69bc1a5a8074adfec876853bbe04a92/packages/react/src/exampleTypes/TagsText.js#L40) and [here](https://github.com/smartprocure/contexture/blob/3658ef6ad69bc1a5a8074adfec876853bbe04a92/packages/react/src/exampleTypes/ExpandableTagsQuery/index.js#L202)

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
dubiousdavid: Let's add `&` to this list and remove it from the list above.

packages/react/src/greyVest/utils.js
```diff
@@ -6,36 +6,47 @@ export let openBinding = (...lens) => ({
...
+let maxCharsPerTagWord = 100
+
+// Strip these characters when splitting a tag into words. We do this to display
+// tags that are closer to what we send to elastic in a `query_string` query.
+//
+// See https://github.com/smartprocure/contexture-elasticsearch/pull/170
+//
+// If in doubt, make a request to the `/{index}/analyze` elasticsearch endpoint
+// to see exactly which characters get stripped out of text.
+let wordRegex = /[^|><!(){}[\]^"~*?\\;,$']+/g
```
dubiousdavid: Let's consolidate this list with the list here: https://github.com/smartprocure/contexture/pull/205/files#diff-69f43c306be976d94d4493c9ca003a660ca8f6084b779bf99ae2288ca58d4519R38

We can pull this out into a common package that's shared between the front end and back end.

dubiousdavid (changes requested):

dubiousdavid:
Let's also make sure to handle the scenario where there is no text left after stripping reserved characters from a tag. In that case, we should remove the tag altogether.

packages/react/package.json
```diff
@@ -68,7 +68,6 @@
     "mobx": "^4.3.1",
     "mobx-react": "^6.3.0",
     "mobx-utils": "^5.0.0",
-    "moment": "^2.24.0",
```
stellarhoof, line 71: This dependency was duplicated.

packages/provider-elasticsearch/src/example-types/filters/tagsQuery.js
stellarhoof: Done

packages/react/src/greyVest/utils.js
stellarhoof: Created the common package and moved the regex there 👍

stellarhoof:
> Let's also make sure to handle the scenario where there is no text left after stripping reserved characters from a tag. In that case, we should remove the tag altogether.

Done

dubiousdavid (approved):
Looks good.

PR #209, opened 2024-04-26
Move off query_string

daedalus28:
Quick example of moving from query_string to match. Needs tests (and all string tests will break).


daedalus28:
Just a note for posterity - this addresses https://github.com/smartprocure/main-application/issues/3807, one of the oldest outstanding issues ever tracked by SmartProcure (first reported on Mar 26, 2018)

PR #210, opened 2024-04-29
ResultTable: HeaderCell and Cell rendering tweaks

stellarhoof:
- Do not wrap `HeaderCell` children in a `span` by default because it makes overriding `HeaderCell` harder.
- Support `headerCellProps` and `cellProps` to accomodate the very common pattern of needing to only pass props to the header/cell components.


packages/react/src/exampleTypes/ResultTable/index.js
```diff
@@ -122,8 +132,17 @@ let ResultTable = ({
...
+                      groupName,
+                      colspan,
+                      headerCellProps,
+                      HeaderCell = Th,
+                      HeaderGroup,
+                    },
+                    j
+                  ) => (
+                    <HeaderCell key={j} colSpan={colspan} {...headerCellProps}>
                       <span>
```
positiveVibes4all: Shouldn't this span be removed also here?

stellarhoof: Indeed. Thanks for catching this 👍

packages/react/src/exampleTypes/ResultTable/Header.js
```diff
@@ -81,136 +90,138 @@ let Header = ({
...
-            <DropdownItem
-              onClick={() => {
-                mutate({ sortField, sortDir: 'desc' })
-              }}
-            >
-              <Icon icon="SortDescending" />
-              Sort Descending
-            </DropdownItem>
-          )}
+        {!disableSort && (
```
conraddavisjr: A request to consolidate the components under a single guard since we're using `{!disableSort &&` twice.

packages/react/src/exampleTypes/ResultTable/Header.js
```diff
@@ -81,136 +90,138 @@ let Header = ({
...
           <DropdownItem
-            onClick={() =>
-              moveColumn(mutate, (i) => i - 1, field, visibleFields, includes)
-            }
+            onClick={() => {
+              mutate({ sortField, sortDir: 'desc' })
+            }}
```
conraddavisjr: Request to simplify
```suggestion
            onClick={() => mutate({ sortField, sortDir: 'desc' })}
```

packages/react/src/exampleTypes/ResultTable/Header.js
stellarhoof: Done

packages/react/src/exampleTypes/ResultTable/Header.js
stellarhoof: Done

conraddavisjr (approved):
There are quite a few things that I'm not in favor of regarding the structure of these components... and I suspect you feel similarly. However, this is an improvement, and one that we should make iteratively, instead of trying to do a huge lift. Thanks for your initiative on this @stellarhoof!

I'd like to get one more review & approval on this from @positiveVibes4all before merging.

positiveVibes4all (approved):
Looks good to go, this does not work with the current spark branch. I tested this on the draft PR in spark linked to this issue and it looked fine.

PR #214, opened 2024-05-14
Enable Export with Excel Format

positiveVibes4all:
## Summary
This PR adds the capability to use an export with '.xlsx' format.

### DIFF
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -35,6 +35,7 @@
     "futil": "^1.76.4",
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15",
+    "read-excel-file": "^5.8.1",
     "write-excel-file": "^2.0.1"
   },
   "packageManager": "yarn@3.3.1"
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -35,7 +35,6 @@
     "futil": "^1.76.4",
     "lodash": "^4.17.21",
     "minimal-csv-formatter": "^1.0.15",
-    "read-excel-file": "^5.8.1",
     "write-excel-file": "^2.0.1"
   },
   "packageManager": "yarn@3.3.1"
--- a/packages/export/package.json
+++ b/packages/export/package.json
@@ -34,7 +34,8 @@
     "contexture-client": "^2.53.7",
     "futil": "^1.76.4",
     "lodash": "^4.17.21",
-    "minimal-csv-formatter": "^1.0.15"
+    "minimal-csv-formatter": "^1.0.15",
+    "write-excel-file": "^2.0.1"
   },
   "packageManager": "yarn@3.3.1"
 }
--- a/packages/export/src/csv.js
+++ b/packages/export/src/csv.js
@@ -40,7 +40,7 @@ export default ({
         recordsWritten = recordsWritten + _.getOr(1, 'recordCount', r)
         await onWrite({ recordsWritten, record: r })
       }
-      await onWrite({ recordsWritten: recordsWritten, isStreamDone: true })
+      await onWrite({ recordsWritten, isStreamDone: true })
       await stream.end()
     })(),
     cancel() {
--- a/packages/export/src/csv.js
+++ b/packages/export/src/csv.js
@@ -11,7 +11,8 @@ export default ({
   // and key of the record.
   // [{ key: string(supports lodash dot notation), label: string, display: function(value, {key, record, transform})}...]
   transform,
-  headers = null, // array of strings to use as headers, array or arrays for multi-line headers
+  //TODO: support multi-line headers in excel and csv when implemented
+  headers = null, // array of strings to use as headers
   onWrite = _.noop, // function to intercept writing a page of records
 }) => {
   stream.write(csv(headers || transformLabels(transform)))
--- a/packages/export/src/csv.js
+++ b/packages/export/src/csv.js
@@ -39,6 +39,7 @@ export default ({
         recordsWritten = recordsWritten + _.getOr(1, 'recordCount', r)
         await onWrite({ recordsWritten, record: r })
       }
+      await onWrite({ recordsWritten: recordsWritten, isStreamDone: true })
       await stream.end()
     })(),
     cancel() {
--- a/packages/export/src/excel.js
+++ b/packages/export/src/excel.js
@@ -2,10 +2,16 @@ import _ from 'lodash/fp.js'
 import F from 'futil'
 import writeXlsxFile from 'write-excel-file/node'
 
-const convertToExcelCell = (value) => {
+let transformLabels = _.map(_.get('label'))
+let maxColumnWidth = 200
+let headerBackgroundColor = '#999999'
+let indexColumnBackgroundColor = '#bbbbbb'
+
+const convertToExcelCell = (value, index) => {
   return {
     wrap: true,
     value: value ? `${value}` : ``,
+    ...((index === 0 && { backgroundColor: indexColumnBackgroundColor }) || {}),
   }
 }
 
@@ -15,9 +21,6 @@ export const writeStreamData = async (stream, writeStreamData) => {
     stream.write(chunk)
   }
 }
-
-let transformLabels = _.map(_.get('label'))
-let maxColumnWidth = 200
 
 export default ({
   stream, // writable stream target stream
@@ -33,14 +36,18 @@ export default ({
 }) => {
   const excelData = [
     _.map(
-      (value) => ({ value, fontWeight: 'bold', backgroundColor: '#999999' }),
+      (value) => ({
+        value,
+        fontWeight: 'bold',
+        backgroundColor: headerBackgroundColor,
+      }),
       headers || transformLabels(transform)
     ),
   ]
 
   let cancel = false
   let recordsWritten = 0
-  let columnWidths = _.map(
+  let columns = _.map(
     (column) => ({ width: column.value.length }),
     excelData[0]
   )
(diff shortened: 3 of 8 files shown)


packages/react/src/FilterButtonList.js
```diff
@@ -10,6 +10,7 @@ import styles from './styles/index.js'
 import { newNodeFromField } from './utils/search.js'
 import { unusedOptions } from './FilterAdder.js'
 import { fieldsToOptions } from './utils/fields.js'
+import { displayLabelFn } from './utils/format.js'
```
stellarhoof: I think these `displayLabelFn` changes snuck in from another branch of yours

packages/export/src/excel.js
```diff
@@ -0,0 +1,73 @@
...
+  transform,
+  headers = null, // array of strings to use as headers, array or arrays for multi-line headers
+  onWrite = _.noop, // function to intercept writing a page of records
+}) => {
+  const excelData = [
+    _.map(
+      (value) => ({ value, fontWeight: 'bold' }),
+      headers || transformLabels(transform)
+    ),
+  ]
```
stellarhoof: Will this work if `headers` is an array of arrays?

packages/export/src/excel.js
```diff
@@ -0,0 +1,73 @@
+import _ from 'lodash/fp.js'
+import writeXlsxFile from 'write-excel-file/node'
+
+const convertToExcelCell = (value) => {
+  return {
+    type: String,
+    typeof: 'string',
```
stellarhoof: What's this `typeof` for?

packages/export/src/excel.js
stellarhoof, line 1: Can you write tests for this file

packages/export/src/excel.js
positiveVibes4all: left over from something I was using for column width, but I abandoned coercing values because already had some odd instances ie: zip codes that were being coerced to strings and numbers because some have a - within them.

I will remove it

packages/export/src/excel.js
positiveVibes4all: What type of test would you suggest? the data is binary, so we really cannot build one like is done for CSV's, I thought about it but it felt like there was little value add for doing this based on this constraint.

packages/react/src/FilterButtonList.js
positiveVibes4all: Possibly, I will remove it. thanks for catching this.

packages/export/src/excel.js
positiveVibes4all: the headers are already being used like this in the CSV file from what I remember, so I did not consider them to be an array of arrays, if that was the case the CSV would not work either from what I read about that library. Can you give an example you have in mind for me to test?

packages/react/src/FilterButtonList.js
positiveVibes4all: There is way more than just that, it all made it in, I will clean up this PR

packages/export/src/excel.js
stellarhoof: I don't have an example I was just going off of the comment for the `headers` argument

> array of strings to use as headers, array or arrays for multi-line headers

packages/export/src/excel.js
stellarhoof: You can read the stream via https://www.npmjs.com/package/read-excel-file, which is nice as you don't need to write a file in the test, keeping it pure.

positiveVibes4all: This seems ok, I am not in love with it because we are using another library to test this one in essence but seems like a start.

stellarhoof: It's by the same author so I think that reduces the risk somewhat

packages/export/src/excel.js
positiveVibes4all: I talked with Kathryn if there is a use case for multi-line headers today and there is not. There could be one if we supported exports of report builder, but that is on the table for discussion. We could support it if the solution we land on winds up including this, but in general for multi header exports we would need a way to support spanning multiple columns and things like this. So for now, I think I will just change this comment to a TODO in both CSV and Excel for the future when we support it from data transfer to export.

packages/export/src/excel.js
positiveVibes4all: I agree, it just feels like we are testing the library itself instead of anything novel about how it is used, I will add it though as it is at least something. Thanks for the feedback on this.

stellarhoof: My thinking is that there is some code that can benefit from testing which is not library related. Another approach is to take a `writeData` callback that looks like

```
      const readStream = await writeXlsxFile(excelData, { columns })
      for await (const chunk of readStream) {
        stream.write(chunk)
      }
```

That way you don't have to read the xlsx file but can still test this function.

packages/export/src/excel.js
```diff
@@ -60,9 +67,7 @@ export default ({
         excelData[1]
       )
       const readStream = await writeXlsxFile(excelData, { columns })
-      for await (const chunk of readStream) {
-        stream.write(chunk)
-      }
+      await writeStreamData(stream, readStream)
```
stellarhoof: There's not much point to this function. My suggestion above was so that you could avoid the call to `writeXlsxFile` in this function and instead pass that + writing the stream as an argument. That way you can write a unit test that passes a dummy function that doesn't use `writeXlsxFile` and you can test the functionality without actually writing an excel file.

positiveVibes4all: I updated the test and code to reflect this.

packages/export/src/csv.js
```diff
@@ -39,6 +40,7 @@ export default ({
         recordsWritten = recordsWritten + _.getOr(1, 'recordCount', r)
         await onWrite({ recordsWritten, record: r })
       }
+      await onWrite({ recordsWritten: recordsWritten, isStreamDone: true })
```
dubiousdavid: You can use the shorthand syntax here: 
```ts
onWrite({ recordsWritten, isStreamDone: true })
```

packages/export/src/excel.js
dubiousdavid: Reminder that this still needs to be done.

packages/export/src/excel.js
```diff
@@ -0,0 +1,75 @@
+import _ from 'lodash/fp.js'
+import writeXlsxFile from 'write-excel-file/node'
+import { writeStreamData } from './utils.js'
+
+const convertToExcelCell = (value) => {
+  return {
+    type: String,
```
dubiousdavid: Is there a reason we are always setting the cell type to string? According to the documentation, the following is true:

_If a cell doesn't have a type, then it is automatically detected from the value, or defaults to a String._

packages/export/src/utils.js
```diff
@@ -1,6 +1,13 @@
 import F from 'futil'
 import _ from 'lodash/fp.js'
 
+export const writeStreamData = async (stream, writeStreamData) => {
```
dubiousdavid: I think this function belongs in /export/src/excel.js. Your test for it lives in /export/src/excel.test.js and it is not used anywhere outside of the excel.js file.

packages/export/src/excel.js
```diff
@@ -0,0 +1,75 @@
...
+        () => ({
+          width: 50,
+        }),
+        excelData[1]
+      )
+      await writeStreamData(
+        stream,
+        async () => await writeXlsxFile(excelData, { columns })
+      )
+      await onWrite({ recordsWritten: recordsWritten, isStreamDone: true })
```
dubiousdavid: You can use the shorthand syntax here: 
```ts
onWrite({ recordsWritten, isStreamDone: true })
```

packages/export/package.json
```diff
@@ -34,7 +34,9 @@
     "contexture-client": "^2.53.7",
     "futil": "^1.76.4",
     "lodash": "^4.17.21",
-    "minimal-csv-formatter": "^1.0.15"
+    "minimal-csv-formatter": "^1.0.15",
+    "read-excel-file": "^5.8.1",
```
dubiousdavid: This package is not being used.

packages/export/src/excel.js
```diff
@@ -0,0 +1,75 @@
...
+              ),
+            transform
+          )
+        )
+        recordsWritten = recordsWritten + _.getOr(1, 'recordCount', r)
+        await onWrite({ recordsWritten })
+      }
+      let columns = _.map(
+        () => ({
+          width: 50,
```
dubiousdavid: What is the default behavior for width? If the default behavior is not smart, we should consider setting the width to the largest value of the following: the length of the column name or 50.

packages/export/src/excel.js
dubiousdavid: Reminder that this still needs to be done.

dubiousdavid (dismissed):

packages/export/src/excel.js
positiveVibes4all: default is not smart, 50 seemed like a reasonable size to start with and wrapping the data set to true.

packages/export/src/excel.js
positiveVibes4all: This is in the current PR and the tests is using it as such.

packages/export/src/excel.js
positiveVibes4all: there are some odd cases like zip code for instance, where some of it was treated as a number and some as a string, this changes the alignment of the values and possibly other underlying things. When I get done with the other changes I can test it again, but this seems safer atm.

packages/export/src/excel.js
positiveVibes4all: I wrote some code to keep track of the longest value and then use the column name as the shortest

packages/export/src/excel.js
positiveVibes4all: I double-checked this, and it handles it fine as it relies on the application to use all the data, I must have originally tested it with some other conditions that are no longer a concern. Updated the code with this change.

dubiousdavid: That's fine. We can circle back to this.

stellarhoof:
Code looks good. Pending approval as I cannot test this with the application due to some bug in the spark PR.

stellarhoof (approved):

PR #215, opened 2024-05-15
Fix tags query tag count when value contains dot

stellarhoof:
Fix issue where the expandable tags query component wouldn't get counts for the tags when the tag values contained dots.


conraddavisjr (approved):
LGTM

dubiousdavid (approved):

PR #218, opened 2024-06-04
Correct merging of highlights

stellarhoof:
Fix issue where highlighting will behave wrongly when the text being highlighted contains substrings that match the highlighting tags (mostly when the highlighted text contains HTML tags).

Refer to the included test for more context.

Here's what the fix looks like in a highlighted document

| **Before** 	| **After** 	|
|:---:	|:---:	|
| Screenshot 2024-06-04 at 4 37 13 PM | Screenshot 2024-06-04 at 4 38 53 PM |

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
+++ b/packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
@@ -53,7 +53,7 @@ export let getNestedPathsMap = (schema, paths) => {
 }
 
 export let stripTags = _.curry((tags, str) =>
-  str.replaceAll(tags.pre, '').replaceAll(tags.post, '')
+  str.replaceAll(getRangesRegexp(tags), '$1')
 )
 
 let getRangesRegexp = _.memoize(
--- a/.changeset/small-phones-clap.md
+++ b/.changeset/small-phones-clap.md
@@ -1,0 +1,7 @@
+---
+'contexture-elasticsearch': patch
+---
+
+More correct merging of highlights
+
+Fix issue where highlighting will behave wrongly when the text being highlighted contains substrings that match the highlighting tags.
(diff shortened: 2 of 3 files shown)


conraddavisjr:
> Fix issue where highlighting will behave wrongly when the text being highlighted contains substrings that match the highlighting tags (mostly when the highlighted text contains HTML tags).
> 
> Refer to the included test for more context.
> 
> Here's what the fix looks like in a highlighted document
> 
> **Before**	**After**
> Screenshot 2024-06-04 at 4 37 13 PM	Screenshot 2024-06-04 at 4 38 53 PM

Love the Before & After layout!

JBezerra:
Hey, I'm willing to test it locally, how can I reproduce the bug? In other words, which document I can face the bug?

stellarhoof:
Here's the ticket: https://govspend.atlassian.net/browse/GS-7046. You will have to change your config to point to production elasticsearch in order to reproduce the bug with the links in that ticket.

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
```diff
@@ -53,7 +53,7 @@ export let getNestedPathsMap = (schema, paths) => {
 }
 
 export let stripTags = _.curry((tags, str) =>
-  str.replaceAll(tags.pre, '').replaceAll(tags.post, '')
+  str.replaceAll(getRangesRegexp(tags), '$1')
```
JBezerra, line 56: Curious to get better understanding on `'$1'`

JBezerra (approved):
Tested locally and working like a charm 🙌🏼

packages/provider-elasticsearch/src/example-types/results/highlighting/util.js
stellarhoof: `$1` refers to the contents of the first capture group in the regexp. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement

JBezerra: Omg you are a mozilla doc wizard, thank you so much!

PR #220, opened 2024-06-12
Let original errors bubble up when running searches

stellarhoof:
It is hard to detect the underlying error that is thrown due to a running searches because we catch them and format them in a few places. This PR removes lets the original errors bubble up, only adding a bit of context to the error message to make it easier to debug where exactly the error happened.

### DIFF
--- a/packages/provider-elasticsearch/src/index.js
+++ b/packages/provider-elasticsearch/src/index.js
@@ -93,23 +93,13 @@ let ElasticsearchProvider = (config = { request: {} }) => ({
 
     let metaObj = { request, requestOptions }
 
-    try {
-      // Log Request
-      node._meta.requests.push(metaObj)
-      let count = counter.inc()
-      debug('(%s) Request: %O\nOptions: %O', count, request, requestOptions)
-      let { body } = await search(request, requestOptions)
-      metaObj.response = body
-      debug('(%s) Response: %O', count, body)
-    } catch (e) {
-      console.error({ e })
-      metaObj.response = e
-      node.error = e.meta.body.error
-      throw {
-        message: `${e}`,
-        ...e.meta.body.error,
-      }
-    }
+    // Log Request
+    node._meta.requests.push(metaObj)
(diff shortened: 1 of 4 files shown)


packages/provider-elasticsearch/src/index.js
```diff
@@ -93,23 +93,13 @@ let ElasticsearchProvider = (config = { request: {} }) => ({
...
-      node._meta.requests.push(metaObj)
-      let count = counter.inc()
-      debug('(%s) Request: %O\nOptions: %O', count, request, requestOptions)
-      let { body } = await search(request, requestOptions)
-      metaObj.response = body
-      debug('(%s) Response: %O', count, body)
-    } catch (e) {
-      console.error({ e })
-      metaObj.response = e
-      node.error = e.meta.body.error
```
stellarhoof, line 107: We don't need this piece of code atm. [Here's the context on why it was added](https://github.com/futil-js/contexture-elasticsearch/pull/145/files#r464778174). However it looks like this code was changed after Sam's PR to throw an error again, making his original intention not work anymore.

packages/server/src/utils.js
```diff
@@ -61,13 +61,8 @@ export let runTypeFunction = (config) => async (name, node, search) => {
...
       : fn(node, schema, config))
   } catch (error) {
-    throw {
-      message: `Failed running search for ${node.type} (${
-        node.key
-      }) at ${name}: ${_.getOr(error, 'message', error)}`,
-      error,
-      node,
-    }
+    error.message = `(at contexture node with key=${node.key} and type=${node.type}) ${error.message}`
```
dshishkov: should we use a different field here, in case message already contains some useful error info we want to see

stellarhoof: The original error message is still preserved at the end. This is just prepending an additional string for context.

dshishkov (approved):

dubiousdavid (approved):

PR #224, opened 2024-06-17
Round the `to` date down to the start of the day

dubiousdavid:
* Set the `to` date to the end of the current day for all `lastXDays` and `lastXMonths` date ranges.
* Set the `from` date to the start of the day for all `lastXDays` date ranges.

See: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/common-options.html#date-math


packages/provider-elasticsearch/src/utils/dateUtil.js
```diff
@@ -35,13 +35,13 @@ export let rangeToDatemath = {
...
   last90Days: { from: 'now-90d', to: 'now' },
   last180Days: { from: 'now-180d', to: 'now' },
-  last12Months: { from: 'now/d-12M', to: 'now' },
-  last15Months: { from: 'now/d-15M', to: 'now' },
-  last18Months: { from: 'now/d-18M', to: 'now' },
-  last24Months: { from: 'now/d-24M', to: 'now' },
-  last36Months: { from: 'now/d-36M', to: 'now' },
-  last48Months: { from: 'now/d-48M', to: 'now' },
-  last60Months: { from: 'now/d-60M', to: 'now' },
+  last12Months: { from: 'now/d-12M', to: 'now/d' },
```
dshishkov: If the `to` date is not supposed to be inclusive, should we be doing the following https://github.com/smartprocure/contexture/blob/main/packages/provider-elasticsearch/src/utils/dateUtil.js#L59 It seems like we are somewhat inconsistent on what each `lastX` uses as the `to` expression

dubiousdavid: I'm actually thinking the following might be the best - `now+1d/d-1ms`. It should cache nicely while also returning results from the current day. The cache expires after six hours, so each time there is a cache miss the query will return any new records. Thoughts?

dshishkov: > I'm actually thinking the following might be the best - `now+1d/d-1ms`. It should cache nicely while also returning results from the current day. The cache expires after six hours, so each time there is a cache miss the query will return any new records. Thoughts?

It makes more sense to me personally to include the current date in the `lastX` expressions, should we adjust the `from` to also do `now+1d/d` in that case so that we are not picking up an extra day in the beginning if the end date include the current day?

stellarhoof: ~Elastic datemath [still gets converted to a timestamp](https://github.com/smartprocure/contexture/blob/main/packages/provider-elasticsearch/src/example-types/filters/date.js#L20-L21) so the cache is ineffective.~

Ignore this. Per our talk, `/d` rounds to the start of the day so we're good 👍 

<details>
<summary>You can see it in the elastic query produced for the issued date filter in spending</summary>
Screenshot 2024-06-17 at 3 54 00 PM
</details>

Also we may want to factor [dateUtil](https://github.com/smartprocure/contexture/blob/main/packages/provider-elasticsearch/src/utils/dateUtil.js) into the new [util package](https://github.com/smartprocure/contexture/tree/main/packages/util) since it looks identical to the one used for other providers.

stellarhoof: @dshishkov In a meeting with Bradley we decided to be inclusive on the entire current day for either past/future ranges. This means that `to` will always be `now+1d/d-1ms` for past ranges and `from` will always be `now/d` for future ranges.

stellarhoof (approved):
Looks good 👍

Also tested locally

dshishkov (approved):

PR #234, opened 2024-06-20
Add missing client types

stellarhoof:
There are lots of new example types in contexture-elasticsearch that had not been accounted for in contexture-client.

### DIFF
--- a/packages/client/src/exampleTypes.js
+++ b/packages/client/src/exampleTypes.js
@@ -360,4 +360,87 @@ export default F.stampKey('type', {
       search: null,
     },
   },
+  dateIntervalGroupStats: {
+    reactors: {
+      groupField: 'self',
+      statsField: 'self',
+      interval: 'self',
+      drilldown: 'self',
+    },
+  },
+  dateRangesGroupStats: {
+    reactors: {
+      groupField: 'self',
+      statsField: 'self',
+      ranges: 'self',
+      drilldown: 'self',
+    },
+  },
+  fieldValuePartitionGroupStats: {
+    reactors: {
+      groupField: 'self',
+      statsField: 'self',
+      matchValue: 'self',
+      stats: 'self',
+      drilldown: 'self',
+    },
+  },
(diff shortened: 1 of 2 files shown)


packages/client/src/exampleTypes.js
```diff
@@ -360,4 +360,87 @@ export default F.stampKey('type', {
       search: null,
     },
   },
+  dateIntervalGroupStats: {
+    reactors: {
+      groupField: 'self',
```
JBezerra, line 365: For my education, looking at the file I usually see `self`, `others` and `all` as valid values, what do they mean?

stellarhoof: These values tell contexture-client what to mark as "dirty" when that property changes

- `self` means mark the node the property belongs to as "dirty"
- `others` means mark nodes that depend on the node the property belongs to as "dirty"
- `all` is a combination of both `self` and `others`

A node is a contexture tree node. For example one of these: https://github.com/smartprocure/contexture/blob/main/packages/client/src/listeners.test.js#L22-L27

If this all sounds confusing then you're not alone. It's not a complicated mental model, we just don't have TypeScript types or good enough documentation in these packages today.

JBezerra: Love it! Could you clarify what do you mean by "dirty", or, when can we consider a node is "dirty"?

stellarhoof: Dirty means that we need to call the backend to get results for that node. The contexture-client library can either submit debounced backend requests when nodes are marked as "dirty" or it can wait for the user to click "Search" to send the request.

JBezerra (approved):
Confirming that by linking to `contexture-client`, it works like a charm.

https://github.com/smartprocure/contexture/assets/20017845/670dff95-d952-4015-ba40-0a9feeeee8bf

stellarhoof:
@JBezerra Thanks for the review!

positiveVibes4all (approved):
Looks good to go!

PR #236, opened 2024-07-02
Add raw example type to contexture client

stellarhoof:
Follow up to https://github.com/smartprocure/contexture/pull/234

### DIFF
--- a/packages/client/src/exampleTypes.js
+++ b/packages/client/src/exampleTypes.js
@@ -443,4 +443,9 @@ export default F.stampKey('type', {
       drilldown: 'self',
     },
   },
+  raw: {
+    reactors: {
+      filter: 'self',
+    },
+  },
 })
--- a/.changeset/smooth-kiwis-shake.md
+++ b/.changeset/smooth-kiwis-shake.md
@@ -1,0 +1,5 @@
+---
+'contexture-client': patch
+---
+
+Add raw example type to client


JBezerra (approved):

packages/client/src/exampleTypes.js
```diff
@@ -443,4 +443,9 @@ export default F.stampKey('type', {
       drilldown: 'self',
     },
   },
+  raw: {
+    reactors: {
+      filter: 'self',
```
daedalus28, line 448: This should probably be others, not self.

stellarhoof: Oh hey welcome! :)

How so? For context, we're using a raw node for a [TermsStatsTable](https://github.com/smartprocure/contexture/blob/main/packages/react/src/exampleTypes/TermsStatsTable.js#L66) and without the client reacting to changes to `filter` the table never refreshes.

daedalus28: `filter` on a `raw` node affects other nodes in the tree by changing its filter. A self reactor tells it to update its own context (rerunning the result function). 

In your example, updating the raw node should update the terms stats node, not itself.

daedalus28: To further clarify - if you’re using `raw` and expecting it’s `filter` property to work like terms stats, you’re likely to be surprised that it’s affecting the entire rest of the search and not just its own context.

PR #238, opened 2024-07-10
Fix default HeaderCell rendering

stellarhoof:
Fix regression introduced in https://github.com/smartprocure/contexture/pull/210 where the header component was being unmounted on every re-render, losing the popover state.

### DIFF
--- a/packages/react/src/exampleTypes/ResultTable/Header.js
+++ b/packages/react/src/exampleTypes/ResultTable/Header.js
@@ -31,6 +31,14 @@ let popoverStyle = {
   width: 'auto',
 }
 
+let DefaultHeaderCell = withTheme(({ children, theme: { Th }, ...props }) => (
+  <Th {...props}>
+    <Flex style={{ display: 'inline-flex', alignItems: 'center' }}>
+      {children}
+    </Flex>
+  </Th>
+))
+
 let Header = ({
   field: fieldSchema,
   includes,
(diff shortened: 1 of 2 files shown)


packages/react/src/exampleTypes/ResultTable/Header.js
```diff
@@ -31,6 +31,14 @@ let popoverStyle = {
...
 }
 
+let DefaultHeaderCell = withTheme(({ children, theme: { Th }, ...props }) => (
+  <Th {...props}>
+    <Flex style={{ display: 'inline-flex', alignItems: 'center' }}>
+      {children}
+    </Flex>
+  </Th>
+))
+
```
conraddavisjr, line 41: Not blocking. I'm leaving this here for posterity. It will be cool, down the road to update the approach to managing components and themes here. Respectfully, it seems like`withTheme` strays away from the benefits that we get from Chakra, and adds a lot of abstraction for simple implementations (like pulling in a Th here)... this typically can lead to less stable implementations.

For my education, why are we using `withTheme` here?

stellarhoof: You're preaching to the choir here :) For the longest time I've wanted to

1. Move contexture-react to spark
2. Get rid of the theme abstraction and just use chakra components

We introduced `withTheme` as part of an initiative to make this library useful for a wide range of users. The theory was that they could plug-in their components using their own UI library via a "theme" and contexture-react would just work. The go-to theme we're using today is called "greyVest" which corresponds to the in-house component library we developed. There's an unstyled theme and a MUI theme experiment as well.

For this bit of code, I'm using `withTheme` to pick up the `Th` component from it.

conraddavisjr (approved):

PR #240, opened 2024-07-17
Fix raw example type reactors

stellarhoof:
Fix regression caused in https://github.com/smartprocure/contexture/pull/236.


JBezerra (approved):
✅

daedalus28 (approved):
Much better 👍

PR #242, opened 2024-07-26
[Facet Fix] - Respect Zero Config While Filtering Correctly 

positiveVibes4all:
## Summary 
This facet filters had a bug that would show all zero counts in Elastic Provider agg buckets even when we were filtering on the field. The expected behavior instead is to show zero counts always but when filtering only the values with counts that match the filter should be shown. This PR corrects this issue, while insuring that the view more button works correctly.

### DIFF
--- a/.changeset/brave-trainers-matter.md
+++ b/.changeset/brave-trainers-matter.md
@@ -1,0 +1,6 @@
+---
+'contexture-elasticsearch': patch
+'contexture-react': patch
+---
+
+Respect include zero node config while filtering properly when filtering is enabled for facets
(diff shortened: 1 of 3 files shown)


JBezerra (approved):
LGTM ✅ 

Confirming it works locally

stellarhoof (dismissed):
There is no need to touch contexture-react, you're just missing the same logic of checking for `node.optionsFilter` [here](https://github.com/smartprocure/contexture/blob/feature/RESPECT_ZEROES_NODE_CONFIG_WITH_NO_FILTER_ONLY/packages/provider-elasticsearch/src/example-types/filters/facet.js#L64).

stellarhoof (approved):
Looks good. Verified locally 👍

PR #244, opened 2024-08-01
Provide more information when throwing errors

stellarhoof:
Use the standard Error class [cause property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause) to add more context about the original elasticsearch error.

<details>
<summary>Before</summary>

```javascript
{
  path: "search"
  method: "create"
  type: "error"
  correlationId: "66ab9da679f959ad29007f5e"
  hook.type: "before"
  message: "(at contexture node with key=undefined and type=fieldValuesGroupStats) x_content_parse_exception: [x_content_parse_exception] Reason: [1:416] [terms] failed to parse field [size]"
  stack: "ResponseError: (at contexture node with key=undefined and type=fieldValuesGroupStats) x_content_parse_exception: [x_content_parse_exception] Reason: [1:416] [terms] failed to parse field [size]\n    at onBody (/Users/ah/Code/smartprocure/spark/node_modules/@elastic/elasticsearch/lib/Transport.js:367:23)\n    at IncomingMessage.onEnd (/Users/ah/Code/smartprocure/spark/node_modules/@elastic/elasticsearch/lib/Transport.js:291:11)\n    at IncomingMessage.emit (node:events:531:35)\n    at endReadableNT (node:internal/streams/readable:1696:12)\n    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)"
}
```

</details>

<details>
<summary>After</summary>

```javascript
{
  path: "search"
  method: "create"
  type: "error"
  correlationId: "66ab9cd459cd7adf705c25ec"
  hook.type: "before"
  message: "x_content_parse_exception: [x_content_parse_exception] Reason: [1:416] [terms] failed to parse field [size]"
  cause: {
    "contexture": {
      "node": {
        "type": "fieldValuesGroupStats",
        "groupField": "Organization.ID",
        "statsField": "LineItem.TotalPrice",
        "size": 0,
        "stats": {
          "sum": true,
          "topHits": {
            "size": 1,
            "_source": [
              "Organization.NameState",
              "Organization.WebSite"
            ]
          }
        },
        "schema": "sp-data-lit"
      }
    },
    "body": {
      "error": {
        "root_cause": [
          {
            "type": "x_content_parse_exception",
            "reason": "[1:416] [terms] failed to parse field [size]"
          }
        ],
        "type": "x_content_parse_exception",
        "reason": "[1:416] [terms] failed to parse field [size]",
        "caused_by": {
          "type": "illegal_argument_exception",
          "reason": "[size] must be greater than 0. Found [0] in [groups]"
        }
      },
      "status": 400
    },
    "statusCode": 400,
    "headers": {
      "content-length": "337",
      "content-type": "application/json",
      "x-cloud-request-id": "r2BGX2psRgmkknQsHE08AQ",
      "x-elastic-product": "Elasticsearch",
      "x-found-handling-cluster": "797f46cce5f64cf7a6ba8ca46dbd154d",
      "x-found-handling-instance": "instance-0000000016",
      "x-opaque-id": "spark-ahernandez@govspend.com",
      "date": "Thu, 01 Aug 2024 14:33:56 GMT"
    }
  }
  stack: "ResponseError: x_content_parse_exception: [x_content_parse_exception] Reason: [1:416] [terms] failed to parse field [size]\n    at onBody (/Users/ah/Code/smartprocure/spark/node_modules/@elastic/elasticsearch/lib/Transport.js:367:23)\n    at IncomingMessage.onEnd (/Users/ah/Code/smartprocure/spark/node_modules/@elastic/elasticsearch/lib/Transport.js:291:11)\n    at IncomingMessage.emit (node:events:531:35)\n    at endReadableNT (node:internal/streams/readable:1696:12)\n    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)"
}
```

</details>

### DIFF
--- a/packages/provider-elasticsearch/src/index.js
+++ b/packages/provider-elasticsearch/src/index.js
@@ -109,7 +109,20 @@ let ElasticsearchProvider = (config = { request: {} }) => ({
     node._meta.requests.push(metaObj)
     let count = counter.inc()
     debug('(%s) Request: %O\nOptions: %O', count, request, requestOptions)
-    let response = await search(request, requestOptions)
+
+    let response
+    try {
+      response = await search(request, requestOptions)
+    } catch (e) {
+      // Provide information about the error in the standard `cause` property
+      // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause
+      e.cause = e.meta ?? {}
+      delete e.meta
+      // Remove duplicated information.
+      delete e.cause.meta
+      throw e
+    }
+
     let body = isAtLeastVersion8(client) ? response : response?.body
 
     // If body has timed_out set to true, log that partial results were returned,
--- a/packages/server/src/utils.js
+++ b/packages/server/src/utils.js
@@ -60,10 +60,13 @@ export let runTypeFunction = (config) => async (name, node, search) => {
     return await (search
       ? fn(node, search, schema, config)
       : fn(node, schema, config))
-  } catch (error) {
+  } catch (e) {
     // Sometimes we throw strings instead of errors
-    const message = _.getOr(error, 'message', error)
-    error.message = `(at contexture node with key=${node.key} and type=${node.type}) ${message}`
+    const error = _.isString(e) ? new Error(e) : e
+    error.cause = {
+      contexture: { node },
+      ...(_.isString(error.cause) ? { error: error.cause } : error.cause ?? {}),
+    }
     throw error
   }
 }
--- a/.changeset/quick-mirrors-tease.md
+++ b/.changeset/quick-mirrors-tease.md
@@ -1,0 +1,6 @@
+---
+'contexture-elasticsearch': patch
+'contexture': patch
+---
+
+Better error reporting
--- a/.changeset/quick-mirrors-tease.md
+++ b/.changeset/quick-mirrors-tease.md
@@ -3,4 +3,4 @@
 'contexture': patch
 ---
 
-Better error reporting
+Use the standard Error class [cause property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause) to add more context about the original elasticsearch error.


ccalbaugh (approved):
Looks like a great improvement!

packages/provider-elasticsearch/src/index.js
```diff
@@ -109,7 +109,20 @@ let ElasticsearchProvider = (config = { request: {} }) => ({
...
+    let response
+    try {
+      response = await search(request, requestOptions)
+    } catch (e) {
+      // Provide information about the error in the standard `cause` property
+      // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause
+      e.cause = e.meta ?? {}
+      delete e.meta
+      // Remove duplicated information.
+      delete e.cause.meta
```
positiveVibes4all, line 122: This is a redundant delete and can be removed.

packages/server/src/utils.js
```diff
@@ -60,10 +60,13 @@ export let runTypeFunction = (config) => async (name, node, search) => {
...
       : fn(node, schema, config))
-  } catch (error) {
+  } catch (e) {
     // Sometimes we throw strings instead of errors
-    const message = _.getOr(error, 'message', error)
-    error.message = `(at contexture node with key=${node.key} and type=${node.type}) ${message}`
+    const error = _.isString(e) ? new Error(e) : e
+    error.cause = {
+      contexture: { node },
+      ...(_.isString(error.cause) ? { error: error.cause } : error.cause ?? {}),
```
positiveVibes4all, line 68: This line causes an error because the precedence of ?? is higher the ?, replace ?? with || to fix this.

[Here is a reference to this precedence in documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#:~:text=The%20nullish%20coalescing%20operator,thrown%20in%20such%20cases.).

positiveVibes4all (changes requested):
Changes requested

packages/provider-elasticsearch/src/index.js
stellarhoof: It's not. `e.meta` is not `e.cause.meta`.

packages/server/src/utils.js
stellarhoof: Did you test this and found an error?

`x ? {} : z ?? {}` and `(x ? {} : z) ?? {}` are equivalent. They both ensure the result of the expression is an object so that the spread doesn't fail.

positiveVibes4all: the nullish operator has a higher precedence, so there is an error in the code when testing this in isolation on runkit. It produces an error when attempting to run because of this, you cannot combine ?? and ? apparently because of the precedence issue. Even with they simple example you posted it still has an error.

packages/provider-elasticsearch/src/index.js
positiveVibes4all: when you set e.cause it clobbers any data and when you delete e.meta it deletes it for cause also  I ran some quick tests to be sure. It is redundant. 
Screenshot 2024-08-05 at 8 53 32 AM

Unless you are expecting meta to have an embedded meta? That does not seem right though.

packages/server/src/utils.js
positiveVibes4all: Actually, I think that runkit has an issue with ?? which makes no sense because the versions look up to date. 

This could be a bad pattern though because of the precedence, but in this case, it is fine because the logic is not problematic.

packages/provider-elasticsearch/src/index.js
stellarhoof: The `error` from elastic doesn't have a `cause` property. Here's a more realistic one

Screenshot 2024-08-05 at 12 08 08 PM

positiveVibes4all: My point is that if you expect meta to have a meta prop then this is fine, but I would not expect that as it makes little sense. If you are solving for that case this is fine as I said above.

positiveVibes4all (approved):

packages/provider-elasticsearch/src/index.js
stellarhoof: > My point is that if you expect meta to have a meta prop then this is fine

Yes that is the expectation

PR #245, opened 2024-08-05
Add new export strategy

stellarhoof:
[Jira ticket](https://govspend.atlassian.net/browse/GS-5732)

Add new export strategy for the contexture-elasticsearch `fieldValuesGroupStats` node.

### Context

We would like to include extra fields for each bucket returned by `terms_stats`, however that node type is deprecated so instead of adding features to it, we decided to add the ability to export `fieldValuesGroupStats` nodes.

### DIFF
--- a/.changeset/shiny-ants-hide.md
+++ b/.changeset/shiny-ants-hide.md
@@ -1,0 +1,5 @@
+---
+'contexture-export': minor
+---
+
+Add a new export strategy for the contexture-elasticsearch fieldValueGroupStats node.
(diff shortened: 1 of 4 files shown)


positiveVibes4all:
Tested and looks good to go

positiveVibes4all (approved):

daedalus28:
FYI - there is already an export strategy for `pivot` which should completely cover this use case and more

dubiousdavid:
> FYI - there is already an export strategy for `pivot` which should completely cover this use case and more

Thanks for looking at the PR!

stellarhoof:
> FYI - there is already an export strategy for pivot which should completely cover this use case and more

Yes we will be using pivot more as we gain more confidence with it. Currently though it's easier for us to use these simpler node types.

conraddavisjr:
Thanks for looking at this PR @daedalus28!
@stellarhoof & @dubiousdavid  
I've created a [follow-up ticket to address this](https://govspend.atlassian.net/browse/GS-7827).

PR #250, opened 2024-08-16
[GS-7591] Add new Step Slider component

JBezerra:
# [GS-7591](https://govspend.atlassian.net/browse/GS-7591)

## Before you start
A few things related to tooling were installed and configured so we can have better DX/reliability going forward:
- ✔️ Typescript
- ✔️ Jest
- ✔️ React testing library

# ⚠️ How to test
In order to have better understanding how to link the changes of this PR to your local spark, please [read the docs](https://www.notion.so/Contexture-4933169365dd48559d794e54bb48b94c).

### On Spark
```
cd web/
yalc link contexture-react
yalc link contexture-client

cd ..
cd api/
yalc link contexture-elasticsearch

cd ..
cd common/
yalc link contexture-client
yalc link contexture-elasticsearch
```

### On Contexture
```
cd packages/react/
yarn prepack && yalc push --force

cd ..
cd provider-elasticsearch/
yarn prepack && yalc push --force

cd ..
cd client/
yarn prepack && yalc push --force
```
### On WebApp
- Pull https://github.com/smartprocure/spark/pull/17007 to your local
- Add `Population Count` filter to your Spending page.

# Demo

https://github.com/user-attachments/assets/48ef5516-8c8a-4a28-ba1c-d163a11f1633

## Context
As a Product Manager, I would like to have a more intuitive way for users to filter using Population Count in spending. Currently it is a range filter where users have to enter in a range to search. It would be nice if we could have pre-set population ranges for users to select and filter.

Start with preset ranges, below are the suggestions:
- 0
- 1,000
- 5,000
- 10,000
- 25,000
- 50,000
- 1,000,000+ ( >= 1,000,000)

# Acceptance Criterias:
- [X] Create new filter type for Population Count in Spending
- [X] Filter type should be a slider with set ranges. Users can drag the slider and it will snap to each range available
- [X] Preset range buckets that slider will snap to as you drag it (component available in chakra)
- [X] Display the selected range above the slider on the left
- [X] Display the value of the range at the start/end of your selection when you hover
- [X] Dynamically calculate the count of records as you make your range selection and display above the slider on the right
- [X] Is it possible to do this without hitting search so users can see the number of expected results when they select a range?


packages/provider-elasticsearch/src/example-types/filters/step.js
conraddavisjr, line 1: I'm curious, what's the LoE to make this TS?
This can be done at a later date and we can discuss as a group if preferred.

packages/provider-elasticsearch/src/example-types/filters/step.js
```diff
@@ -0,0 +1,53 @@
...
+  const max = range[range.length - 1]
+  return { min, max }
+}
+
+const isValueMaxOnSteps = (value, steps) => {
+  const max = steps[steps.length - 1]
+  return value === max
+}
+
+let filter = ({ field, range, steps }) => {
```
conraddavisjr, line 15: 🤔 What do you think about the name `calcRange`?
Looking for something a bit more specific to the function.

packages/provider-elasticsearch/src/example-types/filters/step.test.js
```diff
@@ -0,0 +1,44 @@
...
+describe('step', () => {
+  const input = {
+    type: 'step',
+    field: 'test',
+    range: [0, 500],
+    steps: [0, 500, 1000],
+  }
+
+  it('checks if range is present and not empty', () => {
+    let withoutinput = { type: 'step', field: 'test' }
```
conraddavisjr: 💎 very minor tweak for camelCase
```suggestion
    let withoutInput = { type: 'step', field: 'test' }
```

packages/react/.storybook/main.js
```diff
@@ -50,6 +50,12 @@ export default {
       titlePrefix: 'Live Demos/Index Explorer',
     },
   ],
+  features: { emotionAlias: false },
+  refs: {
+    '@chakra-ui/react': {
+      disable: true,
+    },
+  },
```
conraddavisjr: 🤔 What's the aim of this?

packages/react/src/exampleTypes/StepSlider.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import '@testing-library/jest-dom'
+import userEvent from '@testing-library/user-event'
+import { act, render, screen } from '@testing-library/react'
+
+import StepSlider from './StepSlider'
+
+import testTree from './stories/testTree'
+import { renderWithProviders } from '../utils/test'
+
+describe('StepSlider', () => {
```
conraddavisjr, line 12: 💎 here we _could_ remove the quotes and make it a touch closer to the implementation by not using a string representation.
```suggestion
describe(StepSlider, () => {
```

packages/react/src/exampleTypes/StepSlider.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  beforeEach(() => {
+    renderWithProviders(<StepSlider tree={testTree()} path={['step']} />)
+  })
+
+  it('renders range slider', async () => {
+    const rangeSlider = await screen.getByTestId('chakra-range-slider-track')
+    const rangeSliderThumbs = await screen.getAllByRole('slider')
```
conraddavisjr, line 19: 💎 Minor _potential optimization_ which reuses the query above. From there, I _wonder_ if we'd need to use `await` for rangeSliderThumbs?
```suggestion
    const rangeSliderThumbs = await within(rangeSlider).getAllByRole('slider')
```

packages/react/src/exampleTypes/StepSlider.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(rangeSlider).toBeDefined()
```
conraddavisjr: `toBeDefined()` works. However, in this case I'd like to recommend [toBeInTheDocument()](https://github.com/testing-library/jest-dom/blob/v5.16.5/src/to-be-in-the-document.js).

Reasons:
1. more declarative
2. uses DOM APIs to evaluate the presence of an element

packages/react/src/exampleTypes/StepSlider.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(rangeSliderThumbs).toHaveLength(2)
+  })
+
+  it('renders min and max values', async () => {
+    const rangeLabel = await screen.getByText('Range')
+    const rangeValue = await screen.getByText('0 - 1,000')
+    expect(rangeLabel).toBeDefined()
+    expect(rangeValue).toBeDefined()
```
conraddavisjr: In this context, these `expects` shouldn't be necessary since `getByText` is an assertion and will throw if the element isn't found.

packages/react/src/exampleTypes/StepSlider.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  it('renders results', async () => {
+    const countLabel = await screen.getByText('Results')
+    const countValue = await screen.getByText('400,000')
+    expect(countLabel).toBeDefined()
+    expect(countValue).toBeDefined()
```
conraddavisjr: Same here

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+  Tooltip,
+  Text,
+  HStack,
+  Stack,
+} from '@chakra-ui/react'
+import { contexturify } from '../utils/hoc'
+import { toNumber } from '../utils/format'
+import { FormatterFn, Node, Tree } from '../types/contexture'
+
+/** This theme is an initial step towards looking further
```
conraddavisjr, line 18: ❤️

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+ * at the tokenization to be set by our design system. */
+const theme = {
+  color: {
+    blue: { 100: '#CCE7FF', 500: '#0076DE', 700: '#005199' },
+    gray: { 50: '#FFFFFF', 100: '#E7EAEE' },
+    container: { border: { sm: '#E7EAEE' } },
+    text: {
+      body: {
+        primary: '#495569',
+        tertiary: '#697996',
```
conraddavisjr, line 28: a request to make this secondary.

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+interface StepSliderProps {
+  node: Node<NodeProps, ContextProps>
+  formatter: FormatterFn
+  path: string
+  tree: Tree<NodeProps>
+}
+
+const getDefaultRange = (steps: number[], range: number[]): number[] => {
+  if (range.length === 0) return [0, steps.length - 1]
```
conraddavisjr, line 55: 💎 Suuuper minor, just geeking on pattern for a sec. Thoughts?
```suggestion
  if (!range.length) return [0, steps.length - 1]
```

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const minStep = steps.findIndex((step: number) => step === range[0])
+  const maxStep = steps.findIndex((step: number) => step === range[1])
+  return [minStep, maxStep]
+}
+
+const getRealValue = (steps: number[], value: number): number => {
```
conraddavisjr: A request to elaborate on this name. What is a "real" value in this case, as opposed to one that isn't real?

`getActiveValue` or getCurrentValue; thoughts?

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+  return steps[value]
+}
+
+const StepSlider: FC<StepSliderProps> = ({
+  tree,
+  node,
+  formatter = toNumber,
+}) => {
```
conraddavisjr: 🤔 A request to omit FC, as it implicitly adds props that aren't necessarily accurate to what's intended for the component: https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13


```suggestion
const StepSlider = ({
  tree,
  node,
  formatter = toNumber,
}: StepSliderProps) => {
```

From here, it looks like we may need to account for the case where `toNumber` returns a string, but we're expecting a number.

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const formatValue = (value: number): string => {
+    const realValue = getRealValue(steps, value)
```
conraddavisjr: 🤔 to better understand; what is the intention behind the term "real value"?

conraddavisjr: Is this the "current"/active value by chance?

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const lastStepValue = steps[steps.length - 1]
+    if (realValue === lastStepValue) return `${formatter(realValue)}+`
+    return formatter(realValue).toString()
+  }
+
+  const steps = node.steps
+  const count = node.context.recordsCount ?? 0
+  const defaultRange = getDefaultRange(steps, node.range)
+  const [rangeValues, setRangeValues] = useState(defaultRange)
```
conraddavisjr: 💎 
```suggestion
  const [rangeValues, setRangeValues] = useState<number[]>(defaultRange)
```

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const realFormattedMin = formatValue(rangeValues[0])
+  const realFormattedMax = formatValue(rangeValues[1])
+  const rangeLabel = `${realFormattedMin} -  ${realFormattedMax}`
+  const currentCount = formatter(count).toString()
+
+  return (
+    <Box>
+      <HStack paddingX="4px" justifyContent="space-between">
+        <Label value={rangeLabel} label="Range" />
+        <Label value={currentCount} label="Results" alignment="right" />
```
conraddavisjr: building atop our discussion.
I'd like to propose a minor word tweak
`align`
```suggestion
        <Label value={currentCount} label="Results" align="right" />
```

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+      </HStack>
+      {/** This padding exists because of contexture native margins.
+       * When the slider thumb is encapsulated on the filter UI,
+       * part of it gets cut if no padding is applied. */}
+      <Box marginTop="6px" paddingX="11px">
```
conraddavisjr, line 100: A request for em/rem over px

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  return (
+    <Box>
+      <HStack paddingX="4px" justifyContent="space-between">
+        <Label value={rangeLabel} label="Range" />
+        <Label value={currentCount} label="Results" alignment="right" />
+      </HStack>
+      {/** This padding exists because of contexture native margins.
+       * When the slider thumb is encapsulated on the filter UI,
+       * part of it gets cut if no padding is applied. */}
```
conraddavisjr, line 99: 👍🏾

packages/react/src/exampleTypes/StepSlider.tsx
```diff
@@ -0,0 +1,186 @@
...
+      index={index}
+      borderWidth="2px"
+      background={theme.color.gray[50]}
+      borderColor={theme.color.blue[500]}
+      _hover={{ borderColor: theme.color.blue[700] }}
+      _focus={{ boxShadow: `0 0 0 3px ${theme.color.blue[100]}` }}
+      _active={{ borderColor: theme.color.blue[700] }}
+    />
+  </Tooltip>
+)
```
conraddavisjr: Request for removal of FC and a swap from `px` to em/rem where feasible.

packages/react/src/exampleTypes/TypeMap.js
```diff
@@ -9,6 +9,7 @@ import TagsText from './TagsText.js'
 import Exists from './Exists.js'
 import Bool from './Bool.js'
 import Text from './Text.js'
+import StepSlider from './StepSlider'
```
conraddavisjr, line 12: Does this cause an error?
```suggestion
import StepSlider from './StepSlider.tsx'
```

conraddavisjr (commented):
Outstanding work!

I owe you a review locally.
I was able to do due diligence and managed to leave feedback. Thanks @JBezerra!

packages/provider-elasticsearch/src/example-types/filters/step.js
JBezerra: Even though I would love to do that I'm afraid that's not possible because the object that `step.js` is returning, as well as the other files, is exactly the contract that contexture client is expecting to receive so it can work.

packages/react/.storybook/main.js
JBezerra: Thanks for the question, I will add a comment.

That is because Charka comes with a native storybook automatically assigned to our storybook instance. This tells the storybook to not consider it.

packages/react/src/exampleTypes/StepSlider.test.tsx
JBezerra: That's right! On my first shot I did this but received an unexpected error:
```
It must be a named class, named function, number, or string.
```

That given my constraints, didn't wrestling enough to solve xD

But I’m on It.

JBezerra: Oooh I see why now, `contexturify` is the culprit .

packages/react/src/exampleTypes/StepSlider.test.tsx
JBezerra: Veeery nice! Almost forgot about this. However, unfortunatelly this approach doesn't apply here as `chakra-range-slider-track` and the two sliders are siblings:
```html
<div id="slider-root-5" tabindex="-1" class="chakra-slider css-1kveibn">
  <div data-testid="chakra-range-slider-track" >
    <div id="slider-filled-track-5" class="chakra-slider__filled-track css-1uaoubu"></div>
  </div>
  <div role="slider" tabindex="0"></div>
  <div role="slider" tabindex="0"></div>
</div>

```

packages/react/src/exampleTypes/StepSlider.test.tsx
JBezerra: Perfect!

packages/react/src/exampleTypes/StepSlider.test.tsx
JBezerra: That's correct. But my intention here was to be the more declarative/explicit as possible, even though in this use case that means being redundant.

Another approach would be replacing the `getByText` calls with `queryByText`, so it doesn't throw error but instead returns nullish values. 

Your approach is perfectly valid as well, but I would be concerned that this test case:
```tsx
it('renders min and max values', async () => {
    await screen.getByText('Range')
    await screen.getByText('0 - 1,000')
  })
```

Is less clear than this:
```tsx
it('renders min and max values', async () => {
    const rangeLabel = await screen.queryByText('Range')
    const rangeValue = await screen.queryByText('0 - 1,000')
    expect(rangeLabel).toBeInTheDocument()
    expect(rangeValue).toBeInTheDocument()
  })
```

Due mainly because of the variable name that points to the semantics of the query.

What do you think about that?

packages/react/src/exampleTypes/StepSlider.test.tsx
JBezerra: _I will resolve this comment so we can concentrate the debate on the thread above._

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: Actually that's really the tertiary based on the design system:
Screenshot 2024-08-19 at 10 21 12

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: _mmmm_, I like that... (my time to geek on patterns now) but my brain understands `range.length === 0` faster than `!range. length`. For the second I would have to think _137ms_ more to understand "Oooh! When array is empty` 🤣 

But in the end, feel free to choose one.

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: Right, the "real value" isn't the active or the current ones. What happens is, this Range Slider component is being adapted to become a Step Slider. In order to make that work, we have to "translate" "step" values to its corresponding "real" values. To be more clear:
- Step 0 is 1,000,
- Step 1 is 5,000,
- Step 2 is 10,000
...
- Step 6 is 1,000,000

I will enhance the name but I'm very open to suggestions as well.

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: Love that! Thanks for sharing that article.

About that `From here, it looks like we may need to account for the case where toNumber returns a string, but we're expecting a number.`, not really. 

That's just typescript getting confused with the way `toNumber` was written:
```tsx
export let toNumber = (number, ...params) => {
  if (_.isNumber(number)) {
    let formatter = Intl.NumberFormat(params)
    return formatter.format(number)
  }
  return NaN
}
```

The `NaN` is making the whole typing confused.

I will see what I can do.

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: Type inference is doing that job for us but I will add it so it makes more clear:
Screenshot 2024-08-19 at 10 47 37

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: _I will resolve this thread in favor of concentrating the debate/explanations on the thread above._

packages/react/src/exampleTypes/TypeMap.js
JBezerra: Correct! And to solve that I would need to wrestle a bit with the tooling to understand and solve.
Screenshot 2024-08-19 at 10 56 26

Given my constraints, I would rather skip this fight for now.

conraddavisjr: +1 to this. Sounds good. I'll add a follow-up

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: What do you think about properly discussing that before moving forward with this change? My reason for that consideration is because we are using pixels as our measure everywhere on `contexture` and `spark`.

Doing this change in isolation wouldn't make sense if we are not willing to set that as our new standard from here and beyond. Other than that, that are other considerations we should do before starting using `rem` on a codebase founded in pixels.

packages/react/src/exampleTypes/StepSlider.tsx
JBezerra: _I will resolve this thread in favor of concentrating the debate on the thread above._

packages/react/src/exampleTypes/StepSlider.test.tsx
conraddavisjr: That's an excellent point! 😄. Agreed. I'll resolve this thread.

packages/react/src/exampleTypes/StepSlider.tsx
conraddavisjr: Ahh, for some reason... I was thinking I was looking at a heading. Thanks! Resolving this thread.

packages/react/src/exampleTypes/StepSlider.tsx
conraddavisjr: Good point, I'm happy with either, I like the sleekness of `!range.length`, but I appreciate the explicitness of ===0.

Please leave it as is. Thanks for indulging in my geekiness. 😄

packages/react/src/exampleTypes/StepSlider.tsx
conraddavisjr: What about `em`? What are the repercussions of using that?
I might be missing something, but I don't see drawbacks specifically with em(?).

Also happy to discuss as a team. In fact, that could be more effective to tackle as a group,
v.s a siloed PR. 👍🏾

packages/react/src/exampleTypes/StepSlider.test.tsx
conraddavisjr: Ahhh haha my bad, I was imagining the sliders as children. Thanks!

conraddavisjr (approved):
Verified locally, great work champ!! 🏆

positiveVibes4all (approved):
GTG

PR #255, opened 2024-08-28
[GS-8040] Add open upper bound capability to `step` query

JBezerra:
# [GS-8040](https://govspend.atlassian.net/browse/GS-8040)
## Context
In order to have a controlled way to add/remove an extra step with an open upper bound at the Step Slider component, we need to implement that capability on the `step` query as well.

# ⚠️ How to test
In order to have better understanding how to link the changes of this PR to your local spark, please [read the docs](https://www.notion.so/Contexture-4933169365dd48559d794e54bb48b94c).

### On Spark
```
cd ..
cd api/
yalc link contexture-elasticsearch

cd ..
cd common/
yalc link contexture-client
yalc link contexture-elasticsearch
```

### On Contexture
```
cd ..
cd provider-elasticsearch/
yarn prepack && yalc push --force

cd ..
cd client/
yarn prepack && yalc push --force
```
### On WebApp
- Pull https://github.com/smartprocure/spark/pull/17075 to your local
- Add `Population Count` filter to your Spending page.

# Demo

https://github.com/user-attachments/assets/26aeafa3-9c14-45ca-a1b0-c7f93189314e

### DIFF
--- a/.changeset/hungry-pigs-love.md
+++ b/.changeset/hungry-pigs-love.md
@@ -1,6 +1,5 @@
 ---
 'contexture-elasticsearch': minor
-'contexture-client': minor
 ---
 
 capbility of dealing with `null` ranges added to `step` query on elastic search provider
--- a/.changeset/hungry-pigs-love.md
+++ b/.changeset/hungry-pigs-love.md
@@ -3,4 +3,4 @@
 'contexture-client': minor
 ---
 
-`openUpperBound` feature added to `step` query on elastic search provider
+capbility of dealing with `null` ranges added to `step` query on elastic search provider
--- a/.changeset/hungry-pigs-love.md
+++ b/.changeset/hungry-pigs-love.md
@@ -1,0 +1,6 @@
+---
+'contexture-elasticsearch': minor
+'contexture-client': minor
+---
+
+`openUpperBound` feature added to `step` query on elastic search provider
--- a/packages/provider-elasticsearch/src/example-types/filters/step.js
+++ b/packages/provider-elasticsearch/src/example-types/filters/step.js
@@ -7,39 +7,26 @@ const getMinAndMax = (range) => {
   return { min, max }
 }
 
-const isValueMaxOnSteps = (value, steps) => {
-  const max = steps[steps.length - 1]
-  return value === max
-}
-
-let filter = ({ field, range, steps }) => {
+let filter = ({ field, range }) => {
   const { min, max } = getMinAndMax(range)
-
-  /** In the last step, we do not filter by the upper bound */
-  if (isValueMaxOnSteps(max, steps)) {
-    return {
-      range: { [field]: pickSafeNumbers({ gte: min }) },
-    }
-  }
-
   return {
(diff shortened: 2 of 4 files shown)


.changeset/hungry-pigs-love.md
conraddavisjr, line 1: 🤔 What is "hungry-pigs-love?"

conraddavisjr (approved):
LGTM, thanks @JBezerra!

.changeset/hungry-pigs-love.md
stellarhoof: This is a file auto-generated by [changesets](https://github.com/changesets/changesets) which contains information to update package versions and changelogs once the packages are published.

stellarhoof (changes requested):
With the removal of `steps` from the node, this `step` filter is even closer to the `number` filter so we should consolidate the implementation.

We can do this by adding a flag `calculateRangeTotal` to the `number` node which defaults to `false` (to preserve current behavior). Ex:

```javascript
let result = async(node, search) {
  return F.compactObject({
    // Preserve existing best range functionality
    bestRange: node.findBestRange && (await findBestRange()),
    // Add ability to calculate the range total
    rangeTotal: node.calculateRangeTotal && (await calculateRangeTotal())
  })
}
```

### Update

Ignore my comment above. There is a product requirement for users to be able to switch between the default number component and the slider component and currently that's only supported with different node types so we cannot consolidate `step` and `number` for now.

packages/provider-elasticsearch/src/example-types/filters/step.js
```diff
@@ -7,39 +7,26 @@ const getMinAndMax = (range) => {
...
-
-  /** In the last step, we do not filter by the upper bound */
-  if (isValueMaxOnSteps(max, steps)) {
-    return {
-      range: { [field]: pickSafeNumbers({ gte: min }) },
-    }
-  }
-
   return {
     range: { [field]: pickSafeNumbers({ gte: min, lte: max }) },
```
stellarhoof, line 13: **suggestion:** Delete `getMinAndMax` function.

```suggestion
    range: { [field]: pickSafeNumbers({ gte: range[0], lte: range[1] }) },
```

packages/provider-elasticsearch/src/example-types/filters/step.js
```diff
@@ -7,39 +7,26 @@ const getMinAndMax = (range) => {
...
+        filter: filter({ field, range }),
       },
     },
   }
 }
 
 let result = async (node, search) => {
-  let { field, range, steps } = node
-  const result = await search(buildQuery(field, range, steps))
+  let { field, range } = node
```
stellarhoof, line 28: **suggestion:** Preserve the `number` API

The `number` type uses `min`/`max` instead of `range`.

stellarhoof (approved):

.changeset/hungry-pigs-love.md
conraddavisjr: Ahh copy that, thanks @stellarhoof

PR #257, opened 2024-09-19
[GS-8094] feat: `step` type contract updated to use `min` and `max` instead of …

JBezerra:
# Jira
[GS-8094](https://govspend.atlassian.net/browse/GS-8094)


packages/client/src/exampleTypes.js
```diff
@@ -194,7 +194,8 @@ export default F.stampKey('type', {
     },
     defaults: {
       field: null,
-      range: [0, 1000],
+      min: 0,
+      max: 1000,
```
stellarhoof: I know this was here already but why have a `max` default? I think a range unbounded on its maximum value makes more sense.

packages/client/src/exampleTypes.js
```diff
@@ -194,7 +194,8 @@ export default F.stampKey('type', {
...
       steps: [0, 1000, 5000, 10000, 25000, 50000, 1000000],
```
stellarhoof: **blocker:** Remove default steps

I know this was here before but there's no reasonable default for `steps`. It's totally dependent on the specific data you're querying and it may even be dynamically calculated in the future.

stellarhoof (approved):
- [x] Code looks good.
- [x] Verified locally.

PR #260, opened 2024-10-04
[GS-8482] Revert #259

JBezerra:
# Jira
https://govspend.atlassian.net/browse/GS-8482
# Context
This body of work aims to:
- Revert the major bump merged on #259 and instead, replace it with a minor bump
- Revert the defaults implemented on `step` example type to have a pre-defined `max` value


stellarhoof (approved):
🤞 Code looks good. Let's see if changesets likes this change

PR #263, opened 2025-02-20
[GS-5825]/provide support for typescript_Engineering Ticket

PranitaJainB:
### Jira 
[GS-5825](https://govspend.atlassian.net/browse/GS-5825)

### Requirement
Contexture has never been user friendly, which is compounded by the fact that there is little documentation and no types. This ticket aims to add TypeScript support to it in order to make it more understandable and usable.

NOTE: This ticket is only about enabling TypeScript, not typing the code!

### Verifying

I am testing by renaming a file `packages/client/src/listeners.js` to .ts extension . After that I ran a command `tsc --noEmit` at client folder , and it starts giving Typescript error .


packages/client/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
PranitaJainB, line 1: Hi @stellarhoof , I have made changes into one subfolder named `client`, if this is correct I will do similar changes into other folders(export, provider-elasticsearch, provider-mongo, server, util) too , tailoring the configuration for each folder.

stellarhoof: We should use the same approach used in spark where there is a tsconfig at the root and the individual packages inherit from it.

packages/client/tsconfig.json
```diff
@@ -0,0 +1,55 @@
...
+
+  //compilerOptions inherited from tsconfig.all.jsonc are overridden here.
+  // i have taken reference from spark/web compilerOptions.In that also we are overridding
+  "compilerOptions": {
+    //Provides Node.js global types and  includes types for Vitest which is used by test files
+    "types": ["node", "vitest"],
+
+    //In Spark we are creating dist in 2 folders named 'common' and 'common-node' ,
+    //and those dist files are further consumed by other packages like web etc.
+    //Do we need to create dist folder  for web? "noEmit": true will not create dist folder at all
```
PranitaJainB, line 17: I will be removing comments afterwards.

sandhya-spend (changes requested):
1. `tsx` and `typescript` should be added to `package.json` so that they are setup during install. Also `concurrently`.
2. Also, cannot run certain commands.
Screenshot 2025-02-27 at 1 51 41 PM

PranitaJainB:
> 1. `tsx` and `typescript` should be added to `package.json` so that they are setup during install. Also `concurrently`.
> 2. Also, cannot run certain commands.

Done.

packages/client/tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
+  //Inherits settings from ../tsconfig.all.jsonc which is at root level
+  "extends": ["../../tsconfig.all.jsonc"],
+
+  //Tells TypeScript to only compile files inside the src/ folder and its subdirectories.
+  //It includes .ts, .tsx, .js, and .jsx
+  "include": ["src/**/*"],
+
+  //compilerOptions inherited from tsconfig.all.jsonc are overridden here.
+  // i have taken reference from spark/web compilerOptions.In that also we are overridding
```
sandhya-spend: [Blocker] This is an open-source library, we should not refer to our internal repos/code here.

sandhya-spend (changes requested):
[Blocker[ All the `tsconfig.json` fles are pretty much the same. Consolidate them all to `tsconfig.all.jsonc` and have only the differting options and `extends`.

sandhya-spend (approved):
Thanks. LGTM.

package.json
```diff
@@ -7,6 +7,10 @@
   ],
   "scripts": {
     "foreach": "yarn workspaces foreach --no-private --parallel --verbose",
+    "build:tsc": "tsc --build --verbose",
+    "build": "yarn run build:tsc",
+    "build:watch": "concurrently --raw --kill-others 'tsc --build --verbose --watch --inlineSourceMap'",
```
stellarhoof: No need for `concurrently` as there's only one command being run.

tsconfig.all.jsonc
```diff
@@ -0,0 +1,26 @@
...
+    "noEmitOnError": true, // Prevents output generation if there are TypeScript errors.
+    "incremental": true, // Enables incremental builds for better performance.
+    "skipLibCheck": true, // Skips type checking of library declaration files to speed up compilation.
+    "allowJs": true, // Allows TypeScript to compile JavaScript files.
+    "checkJs": false, // Disables type checking on .js files.
+    "noImplicitReturns": true, // Ensures all functions return a value (e.g., prevents missing return in `if` branches).
+    "noUncheckedIndexedAccess": true, // Forces explicit undefined checks when accessing arrays or objects.
+    "noUnusedParameters": true, // Ensures function parameters are used (reports unused ones).
+    "strict": true, // Enables all strict type-checking options (includes `strictNullChecks`, `noImplicitAny`, etc.).
+    "outDir": "dist" // Specifies the output directory for compiled files.
```
stellarhoof: We'll be using typescript for type-checking only so there is no need to output files. Instead of `outDir` we should specify `"noEmit": true`

tsconfig.all.jsonc
```diff
@@ -0,0 +1,26 @@
...
+    "allowUnusedLabels": false, // Disallows unused labels to avoid confusion in code.
+    "noFallthroughCasesInSwitch": true, // Ensures every switch case explicitly handles its execution flow.
+    "noUnusedLocals": true, // Reports unused local variables to keep the code clean.
+    "noErrorTruncation": true, // Prevents truncation of error messages for better debugging.
+    "isolatedModules": true, // Ensures each module can be transpiled independently for better tooling support.
+    "verbatimModuleSyntax": true, // Enforces explicit use of `type` imports for type-only dependencies.
+    "module": "nodenext", // Uses the ES module system compatible with Node.js.
+    "moduleResolution": "nodenext", // Resolves modules as per Node.js ESM standards.
+    "resolveJsonModule": true, // Allows importing JSON files as modules.
+    "types": ["vite/client", "vitest/globals", "node", "vitest"], // Provides Node.js global types and includes Vitest types.
```
stellarhoof: We don't use `vite` and `node` types are already included by default

```suggestion
    "types": ["vitest/globals"],
```

package.json
```diff
@@ -7,6 +7,10 @@
...
+    "build:clean": "tsc --build --clean && yarn workspaces foreach exec rm -rf dist",
```
stellarhoof: The TypeScript build does not output to the individual `dist` directories so we shouldn't clean them

```suggestion
    "build:clean": "tsc --build --clean",
```

tsconfig.all.jsonc
```diff
@@ -0,0 +1,26 @@
...
+    "jsx": "react", // Specifies JSX transformation method for React applications.
+    "noEmitOnError": true, // Prevents output generation if there are TypeScript errors.
+    "incremental": true, // Enables incremental builds for better performance.
+    "skipLibCheck": true, // Skips type checking of library declaration files to speed up compilation.
+    "allowJs": true, // Allows TypeScript to compile JavaScript files.
+    "checkJs": false, // Disables type checking on .js files.
+    "noImplicitReturns": true, // Ensures all functions return a value (e.g., prevents missing return in `if` branches).
```
stellarhoof: This one we can skip. We've decided not to do it in spark but we haven't removed the setting yet.

stellarhoof (dismissed):

package.json
```diff
@@ -16,6 +20,7 @@
     "@changesets/cli": "^2.26.0",
     "@flex-development/toggle-pkg-type": "^1.0.1",
     "@vitest/coverage-v8": "^2.1.1",
+    "concurrently": "^9.1.2",
```
stellarhoof, line 23: **blocker:** This package is not being used so it can be removed.

package.json
```diff
@@ -25,6 +30,8 @@
     "glob": "^8.1.0",
     "minimist": "^1.2.7",
     "prettier": "^2.8.3",
+    "tsx": "^4.19.3",
```
stellarhoof, line 33: **blocker:** This package is not being used so it can be removed.

stellarhoof:
Some feedback is still outstanding.

PR #268, opened 2025-09-11
Fix deprecated `from` field in `range` query

dubiousdavid:
Contexture uses `range` for the following example types. The `dateRangeFacet` was using the deprecated `from` syntax.

- steps
- date
- dateRangeFacet
- numberIntervalGroupStats
- dateIntervalGroupStats
- dateRangeGroupStats
- percentilesGroupStats
- numberRangesGroupStats
- number


stellarhoof (approved):
Looks good. Thanks for the fix!

PR #274, opened 2025-09-24
Call onUpdateByOthers after results come back

stellarhoof:
We'd like to reset results pagination _after_ results come back, not before.

Additionally, we should also run `onUpdateByOthers` when a field affects both self and others.

### DIFF
--- a/.changeset/dry-lamps-give.md
+++ b/.changeset/dry-lamps-give.md
@@ -1,0 +1,5 @@
+---
+'contexture-client': minor
+---
+
+Run onUpdateByOthers after search results come back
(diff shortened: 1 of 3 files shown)


daedalus28:
Whatever the intention was here, `onUpdateByOthers` exists specifically to handle updates caused by other nodes (and not itself). If I recall, the pivot type for example uses this hook to reset expansions (which you would not want to happen on self updates).

I'm not surprised this was rolled back. I think what you might be looking for is the already existing `onResult` lifecycle hook. Depending on your goal, you could also use `onChange` or `watchNode`. There's also `afterSearch` - tons of existing options.

stellarhoof:
@daedalus28 Hey Sam! There was one fix and one feature mixed in the changes:

- We have a need to run code after the search but only when the node was updated by others.
- We also needed a fix to run `onUpdateByOthers` when the reactor is `all`

We reverted both changes and after talking with product we do need the fix, so I'm opening a PR for it.

Thanks for the insight. Cheers!

PR #280, opened 2025-09-25
Call onUpdateByOthers on properties that have an 'all' reactor

stellarhoof:
Fix `onUpdateByOthers` to run also when a type's reactors is set to `all`. Currently only reactors set to `other` cause this function to run.

### DIFF
--- a/.changeset/wicked-ties-provide.md
+++ b/.changeset/wicked-ties-provide.md
@@ -1,0 +1,5 @@
+---
+'contexture-client': minor
+---
+
+Call onUpdateByOthers on properties that have an 'all' reactor
--- a/packages/client/src/index.js
+++ b/packages/client/src/index.js
@@ -135,11 +135,17 @@ export let ContextTree = _.curry(
         _.find({ path: snapshot(event.path) })
       )(updatedNodes)
       await Promise.all(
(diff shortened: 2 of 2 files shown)


daedalus28:
Just reiterating that `onUpdateByOthers` is intended not to react to self updates from an all reactor - it's about reacting to updates by _others_. I'm pretty sure there was a unit test asserting this behavior at some point. This change is likely to introduce subtle bugs that will be hard to track down (or even notice at first).

Conceptually, `onUpdateByOthers` gives nodes a way to react to things that happen to it that it normally has no control over. By changing the semantics, that no longer exists. It also makes the name confusing because it no longer does what it says - especially when there are existing options in the API that overlap with this change.

If none of the existing hooks work for you, my suggestion would be to introduce a new lifecycle hook to avoid breaking anything existing. Alternatively, rename this one so its name matches its semantics. That would force you to change everything using it and would help surface potential bugs before they happen.

But of course this is just a suggestion. Feel free to ignore it entirely.

stellarhoof:
> Just reiterating that onUpdateByOthers is intended not to react to self updates from an all reactor

It still doesn't do this. The self node is filtered out so `onUpdateByOthers` does not get called on it.

PR #282, opened 2025-09-30
Increase number of highlights for blob

sandhya-spend:
3 is too few highlights, updating the default to 5.


JBezerra:
@sandhya-spend which ticket is this related?

packages/provider-elasticsearch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-elasticsearch",
-  "version": "1.35.0",
+  "version": "1.36.0",
```
stellarhoof: @sandhya-spend The `changeset` package manages all the version bumps so we don't need this change.

sandhya-spend: @stellarhoof undid this.

stellarhoof (approved):

PR #288, opened 2025-12-18
 Hyperlink Support for Excel Exports

PranitaJainB:
# contexture-export

### Overview
This patch introduces a new feature for Excel-type exports where cells can now be rendered as clickable hyperlinks. This is achieved by providing a specific metadata structure that the exporter converts into native Excel formulas.

### New Features
* **Dynamic Excel Hyperlinks**: Cells can now display a custom text label while linking to a specific URL.
* **Metadata-Driven Rendering**: The system identifies links via the `__isHyperlink` flag within the `meta` object.
* **Display Label Mapping**: The `__alias` property is used as the visible display label for the provided URL.
* **Formula Safety**: The exporter automatically handles double-quote escaping to ensure the Excel `HYPERLINK` formula remains valid.

### Data Structure Requirements
To render a cell as a link, the record data for that cell must follow this structure:

```json
{
  "url": "https://example.com/detail/123",
  "meta": {
    "__isHyperlink": true,
    "__alias": "Click to Open Link"
  }
}

### DIFF
--- a/packages/export/src/excel.js
+++ b/packages/export/src/excel.js
@@ -9,10 +9,9 @@ let indexColumnBackgroundColor = '#bbbbbb'
 
 const convertToExcelCell = (value, index) => {
   if (value?.meta?.__isHyperlink) {
-    const excelSafeUrl = value?.url?.replace(/"/g, '""') || ''
-    const displayValue = value?.meta?.__alias || excelSafeUrl || ''
+    const displayValue = value?.meta?.__alias || value?.url || 'Link'
     return {
-      value: `HYPERLINK("${excelSafeUrl}", "${displayValue}")`,
+      value: `HYPERLINK("${value?.url}", "${displayValue}")`,
       type: 'Formula',
       wrap: true,
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/export/src/excel.js [hunk 1]
```diff
@@ -8,6 +8,15 @@ let headerBackgroundColor = '#999999'
 let indexColumnBackgroundColor = '#bbbbbb'
 
 const convertToExcelCell = (value, index) => {
+  if (value?.meta?.__isHyperlink) {
+    const excelSafeUrl = value?.url?.replace(/"/g, '""') || ''
```
positiveVibes4all: ##Blocker 

URLs should never contain " since it’s invalid in a URL. For separation of concerns, this code should assume a valid, well‑formed URL is provided by the call site rather than handling quote escaping here.

positiveVibes4all (approved):
I tested this and it works great; fantastic job getting this over the finish line!

There is, however, a small blocker that needs to be addressed prior to merging. I’m tentatively approving this with the expectation that the blocker is resolved before the merge.

packages/export/src/excel.js
PranitaJainB: url value directly used, cleaning removed. 
Thanks.

PR #291, opened 2026-01-07
Change default elastic highlight tags from b to span

stellarhoof:
Change default elastic highlight tags from b to span


sandhya-spend (approved):
LGTM. Works well. Note that now the highlighted text shows as _italic_ and not **bold**

PR #293, opened 2026-01-23
Support sorting on multiple fields

stellarhoof:
### Summary

Support the ability to sort on multiple fields to the `results` type. This PR deprecates `sortField` and `sortDir` in favor of `sort`

**Before**

```javascript
{
  type: "results",
  sortField: "name",
  sortDir: "asc"
}
```

**After**

```javascript
{
  type: "results",
  sort: [{ field: "name" }, { field: "age", desc: true }]
}
```

### DIFF
--- a/packages/provider-elasticsearch/src/example-types/results/index.js
+++ b/packages/provider-elasticsearch/src/example-types/results/index.js
@@ -2,13 +2,27 @@ import F from 'futil'
 import { getField } from '../../utils/fields.js'
 import { searchWithHighlights } from './highlighting/search.js'
 
+export let getSortParameter = ({ sort, sortField, sortDir }, schema) => {
+  if (sort) {
+    return Object.fromEntries(
+      sort.map(({ field, desc }) => [
+        getField(schema, field),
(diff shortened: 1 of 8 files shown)


packages/provider-elasticsearch/src/example-types/results/index.js
```diff
@@ -2,13 +2,27 @@ import F from 'futil'
 import { getField } from '../../utils/fields.js'
 import { searchWithHighlights } from './highlighting/search.js'
 
+export let getSortParameter = ({ sort, sortField, sortDir }, schema) => {
+  if (sort) {
```
JBezerra: **suggestion:** we could use `!isEmpty` here or something else so we avoid returning an empty object when `sort === {}`.

JBezerra (approved):
- [x] Code looks good
- [x] Verified locally

packages/provider-elasticsearch/src/example-types/results/index.js
stellarhoof: Done

PR #295, opened 2026-02-18
Add today range

yavilagovspend:
### Summary
Adding `today` filter to date range


packages/util/package.json
stellarhoof, line 1: No need to increase the package version manually, you can simply run `yarn changeset` and go from there.

stellarhoof (approved):
This is good to go. Nice job!

PR #297, opened 2026-03-17
ENH: handle exclude and include in highlights

sandhya-spend:
Elasticsearch highlights currently includes all the `node.exclude` fields. Now, the excluded fields are omitted.
Also, `node.include` are globbed for consistency.


packages/provider-elasticsearch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-elasticsearch",
-  "version": "1.38.1",
+  "version": "1.39.0",
```
stellarhoof: There's no need to bump the version manually, changesets will open a PR after this is merged with all the required changes to package.json files.

packages/provider-elasticsearch/src/example-types/results/highlighting/request.js
```diff
@@ -197,14 +197,27 @@ export let getRequestHighlightFields = (schema, node) => {
...
+      const include = expandGlobs(schema, node.include)
+      let subFields = getSchemaSubFields({
+        fields: _.pick(include, schema.fields),
+      })
+      highlightFields = _.pick(
+        _.uniq([...include, ..._.keys(subFields)]),
+        highlightFields
+      )
+    }
+    if (_.isArray(node.exclude)) {
```
stellarhoof: **blocker:** This logic should execute even when `highlightOtherMatches` is true. Otherwise specifying `highlightOtherMatches` will try to highlight fields we excluded explicitly.

sandhya-spend: Done.

stellarhoof (approved):
- [x] Code looks good

